PR #17, opened 2021-11-18
ENH: archived device mixin and tooling

klauer:
## Description
Related:  #14

Most importantly:
* Add an "archived device" ophyd Device translator, which allows you to take a given ophyd Device, turn it into an "archived device", and view its state at any given point in time (assuming it has data inside the archive appliance)
* Add an archiver helper which can retrieve archiver appliance data from multiple appliances in a reasonably performant way

And core things required to get to that point:
* Add a PyepicsPV stand-in/base class that allows for easy control over per-PV data that's represented from the ophyd control layer
* Add tools for rewriting device classes to change `EpicsSignal`-related control layer
* Add basic tooling for custom control layers

Pending:
* Some upstream archiver appliance enhancements will be required for control metadata to be propagated through the `get_snapshot` interface of archapp (https://github.com/pcdshub/archapp/pull/16)
* CI will not pass until https://github.com/pcdshub/archapp/pull/16 is merged

## Motivation and Context
* This will aid with configuration management for automated tests. (#14)
* It will also allow for historical view of ophyd devices in typhos, hutch-python, etc.

## How Has This Been Tested?
* Some unit tests
* Mostly local testing

This PR text and discussions.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,2 +1,3 @@
 [flake8]
 exclude = .git,__pycache__,build,dist,versioneer.py,atef/_version.py,docs/source/conf.py
+max-line-length = 90
--- a/atef/archive_device.py
+++ b/atef/archive_device.py
@@ -1,0 +1,859 @@
+from __future__ import annotations
+
+import copy
+import datetime
+import functools
+import inspect
+import itertools
+import logging
+import numbers
+import time
+from typing import Any, ClassVar, Dict, List, Optional
+
+import aa
+import ophyd
+import pcdsdevices
+# from ophyd import FormattedComponent as FCpt
+from ophyd import Component as Cpt
+from ophyd import Device
+from ophyd import DynamicDeviceComponent as DDCpt
+from ophyd import EpicsSignal, EpicsSignalRO, EpicsSignalWithRBV, Signal
+from ophyd.sim import SynSignal, SynSignalRO
+from ophyd.utils import LimitError
+from pcdsdevices.signal import (EpicsSignalEditMD, EpicsSignalROEditMD,
+                                PytmcSignal, PytmcSignalRO, PytmcSignalRW,
+                                SignalEditMD)
+
+logger = logging.getLogger(__name__)
+
+
+ARCHIVE_CACHE_SIZE = 20_000
+
+
+class ArchiverHelper:
+    _instance_: Optional[ClassVar[ArchiverHelper]] = None
+    pv_to_appliance: Dict[str, aa.fetcher.Fetcher]
+    appliances: List[aa.fetcher.Fetcher]
+
+    def __init__(self):
+        self.pv_to_appliance = {}
+        self.appliances = []
+        self.add_appliance("localhost", 17668)
+
+    @functools.lru_cache(maxsize=ARCHIVE_CACHE_SIZE)
+    def get_pv_at_time(
+        self, pvname: str, dt: datetime.datetime
+    ) -> aa.data.ArchiveEvent:
+        dt = dt.astimezone()  # TODO: aapy raising without this
+        if pvname not in self.pv_to_appliance:
+            for fetcher in self.appliances:
+                try:
+                    event = fetcher.get_event_at(pvname, dt)
+                except ValueError:
+                    ...
+                else:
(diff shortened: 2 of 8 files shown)


klauer:
Before I forget about the implementation details while working on the other parts of atef, want to give this a quick once-over @ZLLentz?

ZLLentz:
I'll take a look

.flake8
```diff
@@ -1,2 +1,3 @@
 [flake8]
 exclude = .git,__pycache__,build,dist,versioneer.py,atef/_version.py,docs/source/conf.py
+max-line-length = 90
```
ZLLentz, line 3: I'm tempted to start bumping this up on other projects too, given that is no longer the 70's

atef/archive_device.py
```diff
@@ -0,0 +1,732 @@
...
+) -> Type[ArchiverDevice]:
+    """
+    Inspect cls and construct an archived device that has the same structure.
+
+    Parameters
+    ----------
+    cls : type[Device]
+        A real Device class to inspect and create an archived Device class
+        from.
+
```
ZLLentz, line 570: some missing parameter docs here

atef/archive_device.py
```diff
@@ -0,0 +1,732 @@
...
+    new_bases = tuple(new_bases)
+
+    clsdict = {}
+    # Update all the components recursively
+    for cpt_name in cls.component_names:
+        cpt = getattr(cls, cpt_name)
+        if "cl" in cpt.kwargs or (
+            inspect.isclass(cpt.cls) and issubclass(cpt.cls, component_classes)
+        ):
+            cpt.kwargs["cl"] = control_layer
```
ZLLentz, line 592: Does this mutate the original Device?

atef/archive_device.py
```diff
@@ -0,0 +1,732 @@
...
+            callback=callback,
+            callback_data=callback_data,
+        )
+
+    def _update_state_from_archiver(
+        self, data: ArchivedValue, as_string=None
+    ) -> Dict[str, Any]:
+        """Update _args and run callbacks based on the ArchivedValue."""
+        if data.value is None:
+            as_string = as_string if as_string is not None else self.as_string
```
ZLLentz, line 708: Is as_string only used when there is no data? That seems wrong to me, unless I misunderstand the intent here.

atef/archive_device.py
klauer: *Thank you* - d'oh

atef/pyepics_compat.py
```diff
@@ -0,0 +1,295 @@
...
+        self, read_access: bool, write_access: bool, *,
+        pv: PyepicsPvCompatibility
+    ) -> NoReturn:
+        ...
+
+
+PyepicsForm = Literal["time", "ctrl", "native"]
+
+
+class PyepicsPvCompatibility:
```
ZLLentz, line 51: I like this implementation detail/approach

.flake8
klauer: It's kinda unacceptable for me to arbitrarily raise it in project xyz without discussing it, but I'm glad you're receptive to it in the end. I was hitting 80 chars on some simple lines with type annotations and struggling to make it not look terrible.

ZLLentz: I don't really mind at all. It's either easy to dial back to 80 with the assistance of tools like `black`, or it isn't easy at all in which case we'd never bother.

ZLLentz: (or, we just accept 90 and move on)

atef/archive_device.py
klauer: Tried to adjust this in 2ce341b; it should indeed be used in both scenarios. I also made some comments regarding how it would be more appropriate to mark it as "disconnected" if not in the archiver, but I have a feeling that will annoy us with broken devices that have a single non-archived PV.

atef/archive_device.py
klauer: 96e0eb8 👍

klauer:
@ZLLentz and I discussed that this project is so preliminary, that it would not hurt to merge in not 100% complete code.

There will be some additional work on this that I'll try to capture in separate issues.

PR #26, opened 2022-02-18
ENH: data reduction implementation for comparisons

klauer:
* Reduce stream of data for a period of time to a single value by way of a defined method
* Run comparison on that value

Relates to https://github.com/pcdshub/pcdsutils/pull/46 ; more code could go there eventually

This is somewhat preliminary, but let's get it merged prior to working on GUI stuff

### DIFF
--- a/atef/check.py
+++ b/atef/check.py
@@ -3,15 +3,14 @@ from __future__ import annotations
 import enum
 import logging
 from dataclasses import dataclass, field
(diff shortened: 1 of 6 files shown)


ZLLentz (approved):
No big comments here, it looks good and useful to me. I'd like to start fiddling with it myself to see how it all works and build out the ui.

PR #27, opened 2022-03-02
REF: rework device/pv config storage

klauer:
* One possibility for a path forward; fleshed out to the point of working tests (lack of discussion time? why not code!)
* Not entirely sold on this just yet
* This represents a "verbose config" storage, avoiding the ugly-but-terse "attribute to checklist" style that was previously here
* Since we've decided on providing a configuration GUI for v0, I myself am at least less worried about making the format something a human would _want_ to edit (though this form absolutely is something they _could_ edit).
* Some is still not implemented (methods to search the configuration for attributes or PVs or tags)

(for discussion tomorrow, hopefully, @ZLLentz !)

### DIFF
--- a/atef/check.py
+++ b/atef/check.py
@@ -604,15 +604,30 @@ class ConfigurationFile:
 
     def get_by_device(self, name: str) -> Generator[DeviceConfiguration, None, None]:
         """Get all configurations that match the device name."""
-        ...
+        for config in self.configs:
+            if isinstance(config, DeviceConfiguration):
+                if name in config.devices:
+                    yield config
 
-    def get_by_pv(self, pvname: str) -> Generator[PVConfiguration, None, None]:
-        """Get all configurations that match the PV name."""
-        ...
+    def get_by_pv(
+        self, pvname: str
+    ) -> Generator[Tuple[PVConfiguration, List[IdentifierAndComparison]], None, None]:
+        """Get all configurations + IdentifierAndComparison that match the PV name."""
+        for config in self.configs:
+            if isinstance(config, PVConfiguration):
+                checks = [check for check in config.checklist if pvname in check.ids]
+                if checks:
+                    yield config, checks
 
     def get_by_tag(self, *tags: str) -> Generator[Configuration, None, None]:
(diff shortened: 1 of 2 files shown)


atef/check.py
```diff
@@ -494,31 +544,35 @@ def _compare(self, value: Number) -> bool:
...
+@dataclass
+class IdentifierAndComparison:
+    #: PV name, attribute name, or test-specific identifier.
+    identifiers: List[str] = field(default_factory=list)
+    #: The comparisons to perform for *each* of the identifiers.
+    comparisons: List[Comparison] = field(default_factory=list)
 
 
 @dataclass
 class Configuration:
```
klauer, line 556: `@serialization.as_tagged_union`

atef/check.py
```diff
@@ -527,48 +581,23 @@ class ConfigurationFile:
     A configuration file comprised of a number of devices/PV configurations.
     """
 
-    #: Dictionary of happi device name to DeviceConfiguration.
-    devices: Dict[str, DeviceConfiguration]
+    configs: List[Union[PVConfiguration, DeviceConfiguration]]
```
klauer, line 584:
```suggestion
    configs: List[Configuration]
```

atef/check.py
```diff
@@ -494,31 +544,35 @@ def _compare(self, value: Number) -> bool:
         return True
 
 
-ItemToChecks = Dict[
-    str,
-    Union[Comparison, Sequence[Comparison]],
-]
+@dataclass
+class IdentifierAndComparison:
```
ZLLentz, line 548: Do these need a `name` field so we can identify them against the other checklist items in `Configuration`

klauer: Added in [4468a7c](https://github.com/pcdshub/atef/pull/27/commits/4468a7c979b5cceff488845ca45af05003d25317)

ZLLentz (approved):
I tried to look through this closely, the parts that I understood all the context for looked good. I think we should roll with this and consider revisiting depending on if issues crop up while making the editor.

klauer:
Sounds good! Clicking the button.

PR #43, opened 2022-04-19
ENH: add the default pydm exception handler

### LINKED ISSUE #42: Set a suitable excepthook for atef config gui
## Expected Behavior

If we raise an exception, don't immediately exit the GUI

## Current Behavior

Undiagnosed buggy behavior causes the UI to crash, losing progress

ZLLentz:
## Description

Add the exception handler from `pydm.exception`, which will display any exceptions we hit in a dialog window instead of simply crashing the app and dumping the exception to terminal.

Use the default handler for now. Later, we may want to expand this similar to what we've done in lucid.

## Motivation and Context

closes #42 

## How Has This Been Tested?

Added a fake error temporarily, pressed the button to cause the error

### DIFF
--- a/atef/bin/config.py
+++ b/atef/bin/config.py
@@ -3,6 +3,7 @@
 """
 import argparse
 
+from pydm import exception
 from qtpy.QtWidgets import QApplication
 
 from ..widgets.config import Window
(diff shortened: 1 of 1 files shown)


klauer (approved):
Working locally (as I think our test suite is lacking coverage at the moment).

image

LGTM 👍

ZLLentz:
Figured it was important to get this one out earlier rather than later to help investigate the other bugs

PR #55, opened 2022-04-22
BUG: Fix issues related to opening files

### LINKED ISSUE #49: Bug: Atef config GUI file opening issues (tree population, freezing)
## Expected Behavior

Open file -> the whole tree is populated

## Current Behavior

Open file -> the tree is only populated as you click on existing elements (each click generates the next layer), at some point it freezes

## Possible Solution

1. Stop lazy loading the GUI elements? Each element is responsible for extending the tree widget slightly.
2. Look for places where we could get stuck in a bad loop

ZLLentz:
## Description

- Fix an issue where we would iterate and extend a list infinitely on startup. I was calling a function meant for adding a new comparison widget instead of just the function meant for setting up an existing comparison.
- Actively create all the ATEF items at startup to fill the whole tree, rather than on an as-needed basis.

## Motivation and Context

closes #49 

## How Has This Been Tested?

Interactively only so far

N/A

### DIFF
--- a/atef/widgets/config.py
+++ b/atef/widgets/config.py
@@ -316,7 +316,7 @@ class AtefItem(QTreeWidgetItem):
     """
     widget_class: type[QWidget]
     widget_args: list[Any]
-    widget_cached: Optional[QWidget]
+    widget_cached: QWidget
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


klauer (approved):
Makes sense based on your issue and proposed fix. Looks good to me 👍

PR #56, opened 2022-04-22
MAINT: switch Severity import to new home, add basic basic test

ZLLentz:
## Description

Follow-up to #48 to fix the import paths for the gui.

Somehow I expected these to be more drastic, I'm not sure what I was thinking.

## Motivation and Context

Code cleanliness: import things from the source, to make sure it doesn't drop out from underneath us. In this case it would have never been a problem but you never know.

More thorough tests are needed for the ui, but this at least makes sure it's importable/runnable on some level.


klauer (approved):
Looks good to me. I agree on the "import from the source", for maintainability and reducing confusion down the line.

PR #57, opened 2022-04-22
ENH: add/finish mechanisms for removing configurations, checklists, and comparisons

### LINKED ISSUE #32: Feature: Implement deletion of list items in atef config gui
## Expected Behavior

I should be able to remove configs, checklists, and comparisons using the GUI.

## Current Behavior

No delete buttons yet.

## Context

Required for GUI completeness.

Related comments:

> I think this todo is still a thing, so issue creating time:
>
> Add a way to delete configurations in the atef config gui.

_Originally posted by @klauer in https://github.com/pcdshub/atef/pull/28#discussion_r853292345_

ZLLentz:
## Description

- Add delete buttons to the top-level overview screen. These are visible for unlocked configurations with empty name fields.
- Make the checklist and comparison delete buttons work properly (by modifying the tree)
- Add docstrings and fill in some more missing docstring info

## Motivation and Context

If we decide that we have too many of any of these, we need to be able to delete them.
closes #32 

## How Has This Been Tested?

Interactively only

Docstrings only

### DIFF
--- a/atef/ui/config_overview_row.ui
+++ b/atef/ui/config_overview_row.ui
@@ -84,6 +84,13 @@
          </widget>
         </item>
         <item>
+         <widget class="QPushButton" name="delete_button">
+          <property name="text">
+           <string>DELETE CONFIG</string>
+          </property>
+         </widget>
+        </item>
+        <item>
          <spacer name="horizontalSpacer">
           <property name="orientation">
            <enum>Qt::Horizontal</enum>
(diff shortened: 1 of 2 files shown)


klauer:
Giving this a try locally - initial reaction is that there probably should be a confirmation dialog prior to deleting.
(If we had a bunch of time, undo/redo would be amazing, but... yeah)

klauer (approved):
I clicked around a bit and didn't find any new bugs, just the one thought above. Code looks fine to me - at least with my understanding of the config GUI code.

ZLLentz:
Right, good call. I think I agree that the two-step (clear text, click button) would be further improved by a confirmation dialog. I wonder if I should short-circuit this in cases where the item is empty... but maybe that's an over-optimization.

Issue #30, opened 2022-04-19
Feature: Implement all Comparison widgets for atef config GUI


ZLLentz:
## Expected Behavior

We should have a widget set up for each comparison subclass in `atef.check` that can be used to modify the dataclass. This widget should be implemented in `atef.widgets.config` as appropriate, expecting a single `QDataclassBridge` as an initialization argument (alongside standard qt arguments like `parent`)

Checklist for how I expect the PRs to be split up:

- [x] Equals/NotEquals
- [ ] ValueSet
- [ ] AnyValue
- [ ] AnyComparison
- [x] Greater/GreaterOrEqual/Less/LessOrEqual
- [x] Range

## Current Behavior

There are only placeholder widgets at this time.

## Context

Required for atef config GUI to be useful.

ZLLentz:
Design ideas:
Equals/Not Equals:
- include a calculated representation of the range defined by the atol/rtol
- include an = or =/= sign as appropriate as a visual indicator
WIP updating comment until I have an idea for each widget

PR #61, opened 2022-04-25
ENH: Equals and NotEquals widgets

ZLLentz:
## Description

An equals/not equals widget for comparison editing. Both are the same widget, just using a different dataclass type and switching out the equals/not equals symbols.

## Motivation and Context

Works toward #30 

## How Has This Been Tested?

Interactively only

Docstrings only

image

### DIFF
--- a/atef/widgets/config.py
+++ b/atef/widgets/config.py
@@ -28,6 +28,7 @@ from ..check import (Comparison, Configuration, ConfigurationFile,
 from ..enums import Severity
 from ..qt_helpers import QDataclassBridge, QDataclassList
 from ..reduce import ReduceMethod
+from ..type_hints import Number, PrimitiveType
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


ZLLentz:
This is nearly to the point I want to take it, just need to work in handling "bool" as a valid primitive type, since the bool -> str and str -> bool translations are not straightforward like for other types.

ZLLentz:
Self-reminder: UI vertical layout spacing is inconsistent here

ZLLentz:
Small detail: the data type defaults to `int` here because the default value for `Equals`'s `value` is an `int`. May consider changing it to a `float`.

ZLLentz:
I think this is as far as I want to take it for now. The widget is responsive and feels reasonably good to use and it's clear what you've entered.

atef/widgets/config.py
```diff
@@ -1209,9 +1211,45 @@ def on_text_changed(self, text: str) -> None:
...
-        font_metrics = self.line_edit.fontMetrics()
-        width = font_metrics.boundingRect(text).width()
-        self.line_edit.setFixedWidth(max(width + 10, 40))
+        match_line_edit_text_width(self.line_edit, text=text)
+
+
+def match_line_edit_text_width(
+    line_edit: QLineEdit,
+    text: Optional[str] = None,
+    min: int = 40,
```
klauer, line 1220: Just a thought:
Maybe 'minimum width in number of characters based on the metrics' would be a bit more intuitive than dealing in pixels here. I think then it'd be like getting the font metrics of a string with `min` `x`s (`"x" * min`)

atef/widgets/config.py
```diff
@@ -1765,10 +1803,319 @@ def __init_subclass__(cls, *args, **kwargs):
...
+        Dataclass bridge to an "Equals" object. This widget will
+        read from and write to the "value", "atol", and "rtol"
+        fields.
+    parent : QObject, keyword=only
+        The normal qt parent argument
+    """
+    filename = 'comp_equals.ui'
     data_type = Equals
+    label_to_type: Dict[str, type] = {
+        'float': float,
```
klauer, line 1848: Hmm, could we instead create an appropriate widget based on the selected data type?
bool -> checkbox, str -> line edit, int/float -> spinbox/double spinbox

atef/widgets/config.py
```diff
@@ -1765,10 +1803,319 @@ def __init_subclass__(cls, *args, **kwargs):
...
+    """
+    Widget to handle the fields unique to the "Equals" Comparison.
+
+    Parameters
+    ----------
+    bridge : QDataclassBridge
+        Dataclass bridge to an "Equals" object. This widget will
+        read from and write to the "value", "atol", and "rtol"
+        fields.
+    parent : QObject, keyword=only
```
klauer, line 1842:
```suggestion
    parent : QObject, keyword-only
```

klauer (commented):
Looking good! Some comments before I take this for a spin.

atef/widgets/config.py
ZLLentz: I was unable to figure out how to do this in terms of character count last time I did this. I could give it another try.

atef/widgets/config.py
ZLLentz: You're right, I should have considered widget types here. Checkbox is absolutely a better interface for the booleans, spinbox is a marginal improvement for the numeric types.

ZLLentz: Actually I'd almost prefer a true/false combobox over a checkbox for the booleans

atef/widgets/config.py
klauer: I think it's unimportant in the scheme of this PR. To my eyes you have everything you need if you know how to use the character metrics to determine a widget width based on a string, though.

atef/widgets/config.py
klauer: I think that is actually clearer - agreed!

atef/widgets/config.py
ZLLentz: I believe you, except that I have not familiarized myself with all of the API and may have made a mistake in not using a monospaced font.

klauer (approved):
Gave this a try locally. It works as advertised and is really nice!
I especially like the intuitive `+-` values and how they update readily with atol/rtol.

ZLLentz:
In the interest of forward progress, I'm going merge and create a new issue for making multiple input widget types. The bits from that will likely be reused across a few of the other comparison widgets.

PR #62, opened 2022-04-25
ENH: happi search tool, ophyd device/component display/selection tool

### LINKED ISSUE #59: Prototype happi-related widgets for component filter/selection
## Current Behavior
In the configuration GUI, you have to type the device name / component name directly and hope you have it correct.

## Proposed solution
Prototype a happi search widget that allows for - in order of priority:

1. Filtering/finding of happi devices, at least by name (some parts may exist already: https://github.com/pcdshub/happi/blob/master/happi/qt/model.py and require customization)
2. Viewing and selecting device components by name
3. Viewing the current state of the device - or via `ArchivedDevice`, at a selected time

This can be prototyped here and moved elsewhere as appropriate. At least (1) can belong in happi itself.

## Context
Configuration GUI discussions.

klauer:
## Description
* Add widget for selection of happi item
* Add widget for selection of component + viewing of live control system status

## Motivation and Context
Closes #59 

## How Has This Been Tested?
Interactively only

Issue + PR

<details>
<summary>Old screenshots</summary>
### By name / list:

image

### By selectable category:
image
</details>

### By name/list (on macOS with simulated IOC)
image

### By beamline (on psbuild)
image

## TODO

- [x] Investigate performance - did enough for now, switched from polling -> user-requested updates
- [x] Add signals for component selection
- [ ] Integrate with config editor
- [ ] Consider if ArchiveDevice can fit into this somehow
- [x] Allow copy/paste of metadata, attribute names, and values
- [x] Look back into sorting bug again
- [x] Add docstrings 
- [ ] Consider adding some tests

### DIFF
--- a/atef/qt_helpers.py
+++ b/atef/qt_helpers.py
@@ -9,9 +9,10 @@ from __future__ import annotations
 import dataclasses
 import functools
 import logging
+import platform
 from typing import Any, Callable, ClassVar, Dict, List, Optional, Tuple, Type
 
-from qtpy import QtCore
+from qtpy import QtCore, QtGui, QtWidgets
 from qtpy.QtCore import QObject
 from qtpy.QtCore import Signal as QSignal
 
@@ -326,3 +327,65 @@ class ThreadWorker(QtCore.QThread):
 def run_in_gui_thread(func: Callable, *args, _start_delay_ms: int = 0, **kwargs):
     """Run the provided function in the GUI thread."""
     QtCore.QTimer.singleShot(_start_delay_ms, functools.partial(func, *args, **kwargs))
+
+
+def get_clipboard() -> Optional[QtGui.QClipboard]:
+    """Get the clipboard instance. Requires a QApplication."""
+    app = QtWidgets.QApplication.instance()
+    if app is None:
+        return None
+
+    return QtWidgets.QApplication.clipboard()
+
+
+def get_clipboard_modes() -> List[int]:
+    """Get the clipboard modes for the current platform."""
+    clipboard = get_clipboard()
+    if clipboard is None:
+        return []
+
+    if platform.system() == "Linux":
+        # Mode selection is only valid for X11.
+        return [
+            QtGui.QClipboard.Selection,
+            QtGui.QClipboard.Clipboard
+        ]
+
+    return [QtGui.QClipboard.Clipboard]
+
+
+def copy_to_clipboard(text, *, quiet: bool = False):
+    """Copy ``text`` to the clipboard."""
+    clipboard = get_clipboard()
+    if clipboard is None:
+        return None
+
+    for mode in get_clipboard_modes():
+        clipboard.setText(text, mode=mode)
+        event = QtCore.QEvent(QtCore.QEvent.Clipboard)
+        app = QtWidgets.QApplication.instance()
+        if app is not None:
+            app.sendEvent(clipboard, event)
+
+    if not quiet:
+        logger.warning(
+            (
+                "Copied text to clipboard:\n"
+                "-------------------------\n"
+                "%s\n"
+                "-------------------------\n"
+            ),
+            text
+        )
+
+
+def get_clipboard_text() -> str:
+    """Get ``text`` from the clipboard. If unavailable or unset, empty string."""
+    clipboard = get_clipboard()
+    if clipboard is None:
+        return ""
+    for mode in get_clipboard_modes():
+        text = clipboard.text(mode=mode)
+        if text:
+            return text
+    return ""
--- a/atef/qt_helpers.py
+++ b/atef/qt_helpers.py
@@ -7,10 +7,14 @@ widgets.
 from __future__ import annotations
 
 import dataclasses
-from typing import Any, ClassVar, Dict, List, Optional, Tuple, Type
+import logging
+from typing import Any, Callable, ClassVar, Dict, List, Optional, Tuple, Type
 
+from qtpy import QtCore
 from qtpy.QtCore import QObject
 from qtpy.QtCore import Signal as QSignal
+
+logger = logging.getLogger(__name__)
 
 
 class QDataclassBridge(QObject):
@@ -271,3 +275,44 @@ class QDataclassList(QDataclassElem):
         self.changed_value.emit(new_value)
         self.changed_index.emit(index)
         self.updated.emit()
+
+
+class ThreadPoolWorker(QtCore.QRunnable):
+    """
+    Worker thread helper.  For running a function in a background QThread.
+
+    Parameters
+    ----------
+    func : callable
+        The function to call when the thread starts.
+    *args
+        Arguments for the function call.
+    **kwargs
+        Keyword arguments for the function call.
+    """
+
+    def __init__(self, func, *args, **kwargs):
+        super().__init__()
+        self.func = func
+        self.args = args
+        self.kwargs = kwargs
+
+    @QtCore.Slot()
+    def run(self):
+        try:
+            self.func(*self.args, **self.kwargs)
+        except Exception:
+            logger.exception(
+                "Failed to run %s(*%s, **%r) in thread pool",
+                self.func,
+                self.args,
+                self.kwargs,
+            )
+
+    @classmethod
+    def new_thread(cls, func: Callable, *args, start=True, **kwargs) -> ThreadPoolWorker:
+        """Start a utility thread in the global QThreadPool."""
+        worker = ThreadPoolWorker(func, *args, **kwargs)
+        if start:
+            QtCore.QThreadPool.globalInstance().start(worker)
+        return worker
--- a/atef/qt_helpers.py
+++ b/atef/qt_helpers.py
@@ -292,7 +292,8 @@ class ThreadWorker(QtCore.QThread):
         Keyword arguments for the function call.
     """
 
-    error_caught = QtCore.Signal(Exception)
+    error_raised = QtCore.Signal(Exception)
+    returned = QtCore.Signal(object)
     func: Callable
     args: Tuple[Any, ...]
     kwargs: Dict[str, Any]
@@ -317,7 +318,9 @@ class ThreadWorker(QtCore.QThread):
                 self.kwargs,
             )
             self.return_value = ex
-            self.error_caught.emit(ex)
+            self.error_raised.emit(ex)
+        else:
+            self.returned.emit(self.return_value)
 
 
 def run_in_gui_thread(func: Callable, *args, _start_delay_ms: int = 0, **kwargs):
--- a/atef/qt_helpers.py
+++ b/atef/qt_helpers.py
@@ -7,6 +7,7 @@ widgets.
 from __future__ import annotations
 
 import dataclasses
+import functools
 import logging
 from typing import Any, Callable, ClassVar, Dict, List, Optional, Tuple, Type
 
@@ -277,7 +278,7 @@ class QDataclassList(QDataclassElem):
         self.updated.emit()
 
 
-class ThreadPoolWorker(QtCore.QRunnable):
+class ThreadWorker(QtCore.QThread):
     """
     Worker thread helper.  For running a function in a background QThread.
 
@@ -291,28 +292,34 @@ class ThreadPoolWorker(QtCore.QRunnable):
         Keyword arguments for the function call.
     """
 
-    def __init__(self, func, *args, **kwargs):
+    error_caught = QtCore.Signal(Exception)
+    func: Callable
+    args: Tuple[Any, ...]
+    kwargs: Dict[str, Any]
+    return_value: Any
+
+    def __init__(self, func: Callable, *args, **kwargs):
         super().__init__()
         self.func = func
         self.args = args
         self.kwargs = kwargs
+        self.return_value = None
 
     @QtCore.Slot()
     def run(self):
         try:
-            self.func(*self.args, **self.kwargs)
-        except Exception:
+            self.return_value = self.func(*self.args, **self.kwargs)
+        except Exception as ex:
             logger.exception(
                 "Failed to run %s(*%s, **%r) in thread pool",
                 self.func,
                 self.args,
                 self.kwargs,
             )
+            self.return_value = ex
+            self.error_caught.emit(ex)
 
-    @classmethod
-    def new_thread(cls, func: Callable, *args, start=True, **kwargs) -> ThreadPoolWorker:
-        """Start a utility thread in the global QThreadPool."""
-        worker = ThreadPoolWorker(func, *args, **kwargs)
-        if start:
-            QtCore.QThreadPool.globalInstance().start(worker)
-        return worker
+
+def run_in_gui_thread(func: Callable, *args, _start_delay_ms: int = 0, **kwargs):
+    """Run the provided function in the GUI thread."""
+    QtCore.QTimer.singleShot(_start_delay_ms, functools.partial(func, *args, **kwargs))
--- a/atef/widgets/core.py
+++ b/atef/widgets/core.py
@@ -1,6 +1,8 @@
 """
 Core classes for atef Qt-based display GUIs.
 """
+from typing import ClassVar
+
 from qtpy.uic import loadUiType
 
 from ..util import ATEF_SOURCE_PATH
@@ -8,7 +10,7 @@ from ..util import ATEF_SOURCE_PATH
 
 class DesignerDisplay:
     """Helper class for loading designer .ui files and adding logic."""
-    filename: str
+    filename: ClassVar[str]
 
     def __init_subclass__(cls):
         """Read the file when the class is created"""
--- a/atef/widgets/core.py
+++ b/atef/widgets/core.py
@@ -1,0 +1,27 @@
+"""
+Core classes for atef Qt-based display GUIs.
+"""
+from qtpy.uic import loadUiType
+
+from ..util import ATEF_SOURCE_PATH
+
+
+class DesignerDisplay:
+    """Helper class for loading designer .ui files and adding logic."""
+    filename: str
+
+    def __init_subclass__(cls):
+        """Read the file when the class is created"""
+        super().__init_subclass__()
+        cls.ui_form, _ = loadUiType(
+            str(ATEF_SOURCE_PATH / 'ui' / cls.filename)
+        )
+
+    def __init__(self, *args, **kwargs):
+        """Apply the file to this widget when the instance is created"""
+        super().__init__(*args, **kwargs)
+        self.ui_form.setupUi(self, self)
+
+    def retranslateUi(self, *args, **kwargs):
+        """Required function for setupUi to work in __init__"""
+        self.ui_form.retranslateUi(self, *args, **kwargs)
--- a/atef/widgets/core.py
+++ b/atef/widgets/core.py
@@ -25,3 +25,20 @@ class DesignerDisplay:
     def retranslateUi(self, *args, **kwargs):
         """Required function for setupUi to work in __init__"""
         self.ui_form.retranslateUi(self, *args, **kwargs)
+
+    def show_type_hints(self):
+        """Show type hints of widgets included in the display for development help."""
+        cls_attrs = set()
+        obj_attrs = set(dir(self))
+        annotated = set(self.__annotations__)
+        for cls in type(self).mro():
+            cls_attrs |= set(dir(cls))
+        likely_from_ui = obj_attrs - cls_attrs - annotated
+        for attr in sorted(likely_from_ui):
+            try:
+                obj = getattr(self, attr, None)
+            except Exception:
+                ...
+            else:
+                if obj is not None:
+                    print(f"{attr}: {obj.__class__.__module__}.{obj.__class__.__name__}")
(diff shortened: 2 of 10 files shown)


ZLLentz:
looking snazzy

klauer:
Getting closer:
image

klauer:
OK, the bones of what I'd like from this tool are here. There's still some TODO items (as in the above PR body), and perhaps more after a preliminary sanity check from @ZLLentz.

ZLLentz:
I'll look through the implementation for high-level understanding + to learn how to run it myself, then I'll also give it a spin

atef/qt_helpers.py
```diff
@@ -271,3 +276,53 @@ def put_to_index(self, index: int, new_value: Any) -> None:
         self.changed_value.emit(new_value)
         self.changed_index.emit(index)
         self.updated.emit()
+
+
+class ThreadWorker(QtCore.QThread):
```
ZLLentz, line 281: Good utility for long-running functions, I like it. We can connect the return signal to an appropriate slot and we'll not freeze the gui to do something intensive.

atef/qt_helpers.py
```diff
@@ -271,3 +276,53 @@ def put_to_index(self, index: int, new_value: Any) -> None:
...
+                self.args,
+                self.kwargs,
+            )
+            self.return_value = ex
+            self.error_raised.emit(ex)
+        else:
+            self.returned.emit(self.return_value)
+
+
+def run_in_gui_thread(func: Callable, *args, _start_delay_ms: int = 0, **kwargs):
```
ZLLentz, line 326: I like this one too. I've occasionally been known to throw out a single shot timer for various reasons and this is a good shortcut.

atef/widgets/core.py
```diff
@@ -0,0 +1,46 @@
...
+Core classes for atef Qt-based display GUIs.
+"""
+from typing import ClassVar
+
+from qtpy.uic import loadUiType
+
+from ..util import ATEF_SOURCE_PATH
+
+
+class DesignerDisplay:
```
ZLLentz, line 11: Good rename :+1:

atef/widgets/core.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(self, *args, **kwargs):
+        """Apply the file to this widget when the instance is created"""
+        super().__init__(*args, **kwargs)
+        self.ui_form.setupUi(self, self)
+
+    def retranslateUi(self, *args, **kwargs):
+        """Required function for setupUi to work in __init__"""
+        self.ui_form.retranslateUi(self, *args, **kwargs)
+
+    def show_type_hints(self):
```
ZLLentz, line 31: This is a good addition for dev too, taking advantage of the copious widget type annotations that are already useful for type checking and IDE completion stuff

atef/widgets/happi.py
```diff
@@ -0,0 +1,369 @@
...
+
+    @property
+    def selected_device_widget(self) -> Union[HappiDeviceListView, HappiDeviceTreeView]:
+        """The selected device widget - either the list or tree view."""
+        if self.radio_by_name.isChecked():
+            return self.happi_list_view
+
+        return self.happi_tree_view
+
+    @QtCore.Slot(str)
```
ZLLentz, line 122: What does an explicit slot decorator do in pyqt?

atef/widgets/happi.py
```diff
@@ -0,0 +1,369 @@
...
+    def client(self, client: Optional[happi.Client]):
+        self._client = client
+        self.happi_tree_view.client = client
+        self.happi_list_view.client = client
+        self.refresh_happi()
+
+
+class HappiItemMetadataView(DesignerDisplay, QtWidgets.QWidget):
+    """
+    Happi item (device) metadata information widget.
```
ZLLentz, line 188: Docstrings are a little light in general in this file- hard for me to immediately understand which thing this is

ZLLentz (commented):
I got hungry before reaching the ophyd widgets- will resume tomorrow. I'm not sure what the best way is to give these a try myself.

I like the way you've approached this

atef/widgets/happi.py
klauer: It's generally unnecessary and more of a personal habit I had when making a signal to annotate the slot. The primary reason would be for disambiguation when the signal itself is overloaded (i.e., multiple call signatures) - in C++ this would be automatically determined by the compiler, but in Python types are just a suggestion after all. According to PyQt5's [documentation](https://www.riverbankcomputing.com/static/Docs/PyQt5/signals_slots.html):

> Although PyQt5 allows any Python callable to be used as a slot when connecting signals, it is sometimes necessary to explicitly mark a Python method as being a Qt slot and to provide a C++ signature for it. PyQt5 provides the [pyqtSlot()](https://www.riverbankcomputing.com/static/Docs/PyQt5/api/qtcore/qtcore-module.html#pyqtSlot) function decorator to do this.
> ... (clipped)
> Connecting a signal to a decorated Python method also has the advantage of reducing the amount of memory used and is slightly faster.
> ...
> It is also possible to chain the decorators in order to define a Python method several times with different signatures. For example:

atef/widgets/happi.py
klauer: I'll work on docs more for sure 👍

atef/widgets/ophyd.py
```diff
@@ -0,0 +1,439 @@
...
+
+            try:
+                changed = np.any(old_value != value)
+            except Exception:
+                ...
+            else:
+                if changed or old_value is None:
+                    for key, value in new_data.items():
+                        setattr(data, key, value)
+                    self.data_changed.emit(attr)
```
ZLLentz, line 184: Does this mean we'll be redrawing the whole table when any value updates?

ZLLentz:
I posted and retracted a question that I realized was wrong
I'll run this a bit now using the snippit you sent me which probably belongs in this thread somewhere.
It's probably worthwhile to add a dev/test/demo entry point for sizable widgets like this.
```
from atef.widgets.happi import *

from typing import List


def test():
    app = QtWidgets.QApplication([])
    client = happi.client.Client.from_config()
    widget = HappiDeviceComponentWidget(client=client)

    def print_selected(items: List[str]):
        print(items)

    widget.item_search_widget.happi_items_selected.connect(print_selected)
    widget.show()
    app.exec_()
    return widget


if __name__ == "__main__":
    test()
```

ZLLentz:
Some issues I had while using the display:
- minor visual bug: if component information has a small number of components, the bottommost component stretches a lot
- slow going to get from click on device -> show metadata
- if the device can't instantiate, we don't even get to see the metadata, current example is the sq1_lamp device
- tons of terminal error spam when we select by non-name keys (for example, stand) but that's probably on happi
- the happi item metadata table is GUI editable via double-clicking on a cell
- "Device selection" is slightly clipped on the n

image

klauer:
Thanks for giving it a go!

Is it obvious that the sections are resizable with the splitters? I think it may only be obvious to me because I know they're there. I wonder if we can do better.

atef/widgets/happi.py
```diff
@@ -0,0 +1,369 @@
...
+        self.list_or_tree_frame.layout().insertWidget(0, view)
+
+        def tree_selection_changed(
+            selected: QtCore.QItemSelection, deselected: QtCore.QItemSelection
+        ):
+            items = [
+                idx.data() for idx in selected.indexes()
+                if idx.parent().data() is not None  # skip top-level items
+            ]
+            self.happi_items_selected.emit(items)
```
ZLLentz, line 106: Is this where we would get the happi info out of this dialog and into a field somewhere in the config gui? What exactly do we get here and why is it a list?

ZLLentz: A list is fine, I just didn't see where the multi-select was currently being used.

Somewhat related, the type information is really opaque with the `QtCore.Signal(list)`. If it's strings that's great, if it's containers that'd also be great, it isn't obvious with my partial understanding.

ZLLentz:
And overall this is pretty great- I'm generally nitpicking the things I ran into but we will make good use of these widgets

ZLLentz:
> Is it obvious that the sections are resizable with the splitters? I think it may only be obvious to me because I know they're there. I wonder if we can do better.

It was pretty clear to me, I just didn't always care to do it when I switching devices a bunch to look at different things. I resized for the first one I looked at.

ZLLentz:
Error spam on shutdown:
```
Poll thread for mfx_dia_ipm.diode.state.motor.high_limit_travel @ 1.000 sec failure
  File "/cds/home/z/zlentz/pydev/atef/widgets/ophyd.py", line 148, in _update_attr
    setpoint = data.signal.get_setpoint()
  File "/cds/home/z/zlentz/miniconda3/envs/dev/lib/python3.9/site-packages/ophyd/signal.py", line 1613, in get_setpoint
    info = self._get_with_timeout(
  File "/cds/home/z/zlentz/miniconda3/envs/dev/lib/python3.9/site-packages/ophyd/signal.py", line 1127, in _get_with_timeout
    info = pv.get_with_metadata(
  File "/cds/home/z/zlentz/miniconda3/envs/dev/lib/python3.9/site-packages/epics/pv.py", line 45, in wrapped
    raise RuntimeError('Expected CA context is unset')
RuntimeError: Expected CA context is unset
```

ZLLentz:
And last things, I promise:
- There's a small pause after selecting the happi name where the GUI becomes unresponsive. I suspect there's more that needs to be stuffed into the QThread.
- The GUI seems to also big freeze on first EPICS updates of large devices like at2l0. Try running the scrollbars up and down at various points, whenever you can't move the scrollbar the main qt thread is being fully used by something. Another way to verify a freeze is to change the window size- the GUI will not adjust the size of widgets until after the freeze is over.

klauer:
Performance is one of my key concerns here, for sure. I'm very open to input on a path to improve it.

> There's a small pause after selecting the happi name where the GUI becomes unresponsive. I suspect there's more that needs to be stuffed into the QThread.

... but at the same time, I think we may be running into GIL issues somewhere along the way. Some of pyepic's stuff has a bunch of `libca.poll()` statements in there for example. Ideally, we could shove this into a subprocess and just get device information back/forth that way. It's a different can of worms with that though.

ZLLentz:
I don't think you need to push the performance to the limit in this PR's scope- it's totally possible that as-is will be perfectly good for the purposes of the config gui. If not, we can bully pyepics until it accepts our demands some other time.

atef/widgets/happi.py
klauer: Agreed on the opaqueness. 

I initially had the `Signal` with `List[str]`, but PyQt doesn't want annotations, it wants objects. 🤷‍♂️ 
I wonder if either pyqt or pyside will eventually have better support in this regard.

atef/widgets/happi.py
```diff
@@ -0,0 +1,369 @@
...
+        self._client = None
+        self._device_worker = None
+        self._device_cache = {}
+        self.client = client
+        # self._setup_ui()
+        self.item_search_widget.happi_items_selected.connect(
+            self._new_item_selection
+        )
+
+    def _new_item_selection(self, items: List[str]):
```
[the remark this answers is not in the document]
klauer: Resolved in 19ea0e0

atef/widgets/happi.py
ZLLentz: Ugh, qt probably wants you to use QStringList

klauer: Ahh, true - I totally forgot about that...

klauer: Wait, QStringList doesn't exist in PyQt5 any more! It went away with `QString`, thankfully. I guess I'll stick with `list` and a comment about the annotation.

klauer:
OK, this is in a lot better of a spot than yesterday.

* New signals for the test:
```
    widget.item_search_widget.happi_items_selected.connect(print_selected)
    widget.device_widget.attributes_selected.connect(print_selected)
```

* New select button
* Context menus for copying data
* Context menus for selecting attributes
* Button for selecting attributes (multiple are OK)
* Switched to user-requested updates of data (not polling by default, though there are spots in there for it to be customized later)
* Speed is marginally improved, but not great
* Text filters for devices

New screenshot - will push this in the PR description tomorrow as it's past 5PM and I'm tired:
image

klauer:
It wasn't obvious to me, but the sorting in the GUI is user-configurable when clicking the tree's horizontal header:
image

So I'm marking the sorting bug as resolved for now.

I think I'd like to move forward with the provided signals and see how we can best integrate this with the config GUI, @ZLLentz. I'll request another review and we can chat about any more of the details.

Some random thoughts:
* I think this will work well for adding new devices to the list and selecting one or more components by name (the signals for both are noted above).
* I don't think there are enough hooks to make this convenient for an editing scenario, that is:
    * We have a configuration file already
    * There are one or more devices already configured with a checklist
    * The user wants to modify the device list, go check happi metadata, or tweak the selected components
    * I could imagine a "jump to device in the happi dialog" or "jump to component in the happi/ophyd dialog" as part of the regular configuration GUI
* This won't work well just yet for another weird scenario:
   * Given two or more _similar but not equivalent_ devices which share some component names
   * Show the user only the components that can be used for _all_ of the devices
   * This may not be as weird as it seems at first glance - one device may be a subclass of another, etc.

ZLLentz:
There's a small wrinkle in that in the configuration we don't save the devices and components in the same place. That is, we select some number of "devices" per device config, and then we add some number of "checklists", and each checklist has some number of "signals". So the config implementation will probably use multiple variants of this- one that outputs selected device names, while displaying the metadata, and another that looks exactly like your screenshots for selecting the components, but with a device subset potentially? Details slightly unclear.

I'll review the code and pay close attention to the various widget types and their signals

atef/widgets/happi.py
ZLLentz: Oh thank goodness I never want to see it again

atef/widgets/happi.py
```diff
@@ -0,0 +1,470 @@
...
+from qtpy.QtWidgets import QWidget
+
+from ..qt_helpers import ThreadWorker, copy_to_clipboard
+from .core import DesignerDisplay
+from .ophyd import OphydDeviceTableWidget
+
+logger = logging.getLogger(__name__)
+
+
+class HappiSearchWidget(DesignerDisplay, QWidget):
```
ZLLentz, line 24: This widget looks :+1: to me, I like the implementation. In isolation this behaves pretty snappily with just a one-time startup cost for reading the happi database.

In the config gui, I'm imagining:
- instead of adding an empty line edit, the add device button would open a dialog
- this dialog would include the HappiSearchWidget and probably the metadata widget
- the confirm/close options at the bottom of the dialog would close the dialog and add an entry of the current selection on confirm

atef/widgets/happi.py
```diff
@@ -0,0 +1,470 @@
...
+        self.edit_filter.textEdited.connect(set_filter)
+        view.proxy_model.setRecursiveFilteringEnabled(True)
+
+    def _tree_view_context_menu(self, pos: QtCore.QPoint) -> None:
+        """Context menu for the happi tree view."""
+        self.menu = QtWidgets.QMenu(self)
+        index: QtCore.QModelIndex = self.happi_tree_view.indexAt(pos)
+        if index is not None:
+            def copy(*_):
+                copy_to_clipboard(index.data())
```
ZLLentz, line 151: The clipboard stuff here will be helpful in future contexts outside the scope of the config widget

atef/widgets/happi.py
```diff
@@ -0,0 +1,470 @@
...
+
+    @client.setter
+    def client(self, client: Optional[happi.Client]):
+        self._client = client
+        self.happi_tree_view.client = client
+        self.happi_list_view.client = client
+        self.refresh_happi()
+
+
+class HappiItemMetadataView(DesignerDisplay, QtWidgets.QWidget):
```
ZLLentz, line 243: This widget looks good too. My understanding is that the runtime intended usage is to set `item_name` to whichever device name I'd like to display the metadata for. Pretty solid.
We'll probably use this to provide much needed context information when the user is selecting devices.

atef/widgets/happi.py
```diff
@@ -0,0 +1,470 @@
...
+        self,
+        parent: Optional[QWidget] = None,
+        client: Optional[happi.Client] = None,
+    ):
+        super().__init__(parent=parent)
+        self._client = None
+        self._device_worker = None
+        self._device_cache = {}
+        self.client = client
+        # self._setup_ui()
```
ZLLentz, line 412:
```suggestion
```

atef/widgets/happi.py
```diff
@@ -0,0 +1,470 @@
...
+        self._item_name = item_name
+        self._update_metadata()
+
+    @property
+    def metadata(self) -> Dict[str, Any]:
+        """The current happi item metadata, as a dictionary."""
+        return dict(self._metadata)
+
+
+class HappiDeviceComponentWidget(DesignerDisplay, QWidget):
```
ZLLentz, line 375: This would be a pretty fun one to have a demo entry point for

atef/widgets/happi.py
```diff
@@ -0,0 +1,470 @@
...
+    """
+    Happi item (device) search widget + component view and selection.
```
ZLLentz, line 377: For the config gui there must be a way to use this wholesale and show it with a pre-filter on the device list to only include the devices available in the config.

I'm tempted to ignore the problems related to non-uniform signal sets for now.

atef/widgets/ophyd.py
```diff
@@ -0,0 +1,729 @@
...
+
+    Parameters
+    ----------
+    device : ophyd.Device
+        The ophyd device to poll.
+
+    poll_rate : float
+        The poll rate in seconds. A zero or negative poll rate will indicate
+        single-shot mode.  In "single shot" mode, the data is queried exactly
+        once and then the thread exits.
```
ZLLentz, line 101: I think this is a good decision (poll rate > monitor, allow single shot). Too many devices have fast-updating PVs and it's not the purpose of this GUI to be a live control system viewer.

atef/widgets/ophyd.py
```diff
@@ -0,0 +1,729 @@
...
+
+            if self.poll_rate <= 0.0:
+                # A zero or below means "single shot" updates.
+                break
+
+            elapsed = time.monotonic() - t0
+            time.sleep(max((0, self.poll_rate - elapsed)))
+
+
+class PolledDeviceModel(QtCore.QAbstractTableModel):
```
ZLLentz, line 248: Absolutely worth pulling this and the other ophyd Qt stuff out into a more central module when we're done trialing them here in atef. :+1:

atef/widgets/ophyd.py
```diff
@@ -0,0 +1,729 @@
...
+
+        if new_model:
+            # Only start an update if we haven't previously gotten information
+            # about the device
+            model.start()
+
+
+class OphydDeviceTableWidget(DesignerDisplay, QtWidgets.QFrame):
+    """
+    A convenient frame with an embedded OphydDeviceTableView.
```
ZLLentz, line 634: What exactly is this used for over just having an OphydDeviceTableView?

ZLLentz (approved):
I think this is really solid. It'd hard to find anything to take fault in here. I imagine that we'll have to start using it to understand if there are any hidden quirks in how the API is set up, but from where I'm sitting I think I understand how I'd use these to make the user's device/component selection in the config GUI much smoother.

Also, the snippit you sent me for testing may as well be a `happi` built-in. Maybe `happi browse` or something like that.

atef/widgets/happi.py
klauer: That sounds good from the config gui perspective. I think keeping the happi search widgets + metadata widget (and explicitly not the component widget) makes sense.

atef/widgets/happi.py
klauer: I took this out of my pydm PR and fixed it up further here. Agreed - it probably should go in pcdsutils.qt or something.

atef/widgets/happi.py
klauer: You've got the right idea, intended usage-wise. It may have more uses than just what we're considering at the moment, even. Happi metadata editor utility instead of our poor excuse for an update CLI?

atef/widgets/happi.py
klauer: The filter should support regular expressions, so the easiest way would probably be:
reach in to the instantiated dialog's filter textbox and call `setText(regex)`.
It wouldn't cause them to be selected, but I guess it's better than nothing?

atef/widgets/ophyd.py
klauer: `OphydDeviceTableView` is _just_ the table view, but this widget `OphydDeviceTableWidget` includes the buttons and filters that surround it. 

The goal in my head was to allow piece-meal combining of the table view into new widgets, but also give a conveniently-instantiable widget that does probably everything you'd want.

klauer:
> Also, the snippit you sent me for testing may as well be a happi built-in. Maybe happi browse or something like that.

I think that would be great. Having a front-end for searching happi, inspecting metadata, looking at devices, and opening up typhos screens even would be pretty neat...

Thanks for the thoughtful feedback here, @ZLLentz. I think we're probably good to move forward.

I'll take another look at this thread tomorrow and see if we need to create any issues.

PR #66, opened 2022-04-27
ENH: GreaterWidget, LessWidget, OrEquals variants

ZLLentz:
## Description

Add config widgets for the following data classes:
- `Greater`
- `GreaterOrEquals`
- `Less`
- `LessOrEquals`

Includes some code consolidation/deduplication for all of these very similar classes.

Compared to the others, I think these were/will be the simplest to define- there is only one field to represent.

## Motivation and Context

Progress towards #30

## How Has This Been Tested?

Interactively only

Docstrings only

image

### DIFF
--- a/atef/ui/comp_equals.ui
+++ b/atef/ui/comp_equals.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>238</width>
-    <height>97</height>
+    <height>113</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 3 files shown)


ZLLentz:
Moderately concerned that I went too crazy with the class inheritance
But this should handle the greater/less widgets and re-use code among all these super simple widgets

klauer (approved):
I see you were concerned about too much inheritance - 
I think your implementation justifies it here. There's so much code re-use and conceptual similarity between them, of course, being inequality/equality checks.

Nothing stood out as problematic to me. I would like to get a chance to try this out soon.
I think it's in a perfectly mergeable state now.

PR #67, opened 2022-05-03
ENH: Range widget

ZLLentz:
## Description

Add a widget to manipulate the Range comparison dataclass while providing a helpful visualization of the range.
The visualization will change its colors and update the relative spacing of the thresholds as the user changes them.

I've implemented everything I could think of for the visualization while keeping it simple. Some ideas that I discarded because they sounded annoying:
- logarithmic ranges
- single QPainter instead of multi widgets in a grid
- use the mathematical filled/empty dots on the bounds to show inclusive/not

## Motivation and Context

Continues #30 
The visualization is motivated by a need to disambiguate what exactly the behavior of the checks will be as we manipulate the values here.

## How Has This Been Tested?

Interactively only

Docstrings only

First open:
image

Sensible settings:
image

The borders move:
image

Inclusive:
image

Inversion:
image

### DIFF
--- a/atef/widgets/config.py
+++ b/atef/widgets/config.py
@@ -2388,16 +2388,12 @@ class RangeWidget(CompMixin, DesignerDisplay, QWidget):
             # Something is still None
             ordered = False
         real_space = self.width() * 0.7
-        if ordered:
-            # Looks OK, show everything
-            self.left_yellow_line.show()
-            self.right_yellow_line.show()
-            self.vertical_line_2.show()
-            self.vertical_line_3.show()
-            self.warn_low_label.show()
-            self.warn_high_label.show()
-        else:
-            # Not OK, hide warnings, scale green, set bound colors, and end
+
+        if not ordered or self.bridge.invert.get():
+            # No warning bounds, something is nonphysical, or we are inverted
+            # Note: inversion implies a nonsensical "fail and warn" region
+            # that should be ignored.
+            # Hide warnings, scale green, set bound colors, and end
             self.left_yellow_line.hide()
             self.right_yellow_line.hide()
             self.vertical_line_2.hide()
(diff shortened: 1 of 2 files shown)


ZLLentz:
There are some sizing issues if the user puts in a really really big number- not sure how important these would be

ZLLentz:
I'm also now realizing before I skip off to a late lunch- inversion coloring might not be correct for the warning ranges, I'm not 100% sure of the implementation

klauer:
> There are some sizing issues if the user puts in a really really big number- not sure how important these would be

Not important, in my opinion.

> I'm also now realizing before I skip off to a late lunch- inversion coloring might not be correct for the warning ranges, I'm not 100% sure of the implementation

Inverting with warning ranges is a bit too confusing, if you ask me. Ignoring the inclusive/exclusive setting:
With non-inverted logic:
  * No warning limits set
    * If below `low` or above `high` -> `severity_on_failure`
    * Value falling within `low`/`high` -> success
  * Warning limits set
    * If below `low` or above `high` -> `severity_on_failure`
    * Warn if value is in `(low, warn_low)` or `(warn_high, high)`. 
    * Value falling within `warn_low`/`warn_high` -> success


With inverted logic:
  * No warning limits set
    * If value between `low` and `high` -> `severity_on_failure`
    * Value falls outside `low` and `high` -> success
  * Warning limits set
     * Invert the logic above for each step; I don't think it's sensible

klauer (approved):
I think this looks really great and it should be rather intuitive for the end-user. I plan on playing around with this more in a bit. That shouldn't hold up merging this PR.

atef/widgets/config.py
```diff
@@ -2210,3 +2213,220 @@ class LessOrEqualWidget(CompMixin, GtLtBaseWidget):
...
+        """
+        # Cute trick: swap red and green if we're inverted
+        if self.bridge.invert.get():
+            green = self.red
+            red = self.green
+        else:
+            green = self.green
+            red = self.red
+        yellow = self.yellow
+        self.left_red_line.penColor = red
```
klauer, line 2359: I think the vertical lines could use a larger pen width to make them more visible

ZLLentz: Easy to adjust

ZLLentz:
> Inverting with warning ranges is a bit too confusing

I think I agree- should I just hide the warning thresholds under inversion?

atef/widgets/config.py
ZLLentz: Thicker lines: 
image

ZLLentz:
I'm going to remove the yellow region under inversion, much like how the gui displays with no warning threshold set, to make it clear that "fail with warning" is not a sensible outcome for a test

PR #70, opened 2022-05-03
REF/ENH: hook up device list with the happi search tool

klauer:
## Description
* Use the happi search widget as a device selector
* Use the device/component widget as an attribute selector
* Select happi items and double-click them or click "choose"/'select" with them selected

## Motivation and Context
* Typing in device names is not fun and error-prone
* Typing in component names is even worse

## How Has This Been Tested?
* Interactively

* PR text, as we do 

image

image

### DIFF
--- a/atef/ui/config_group.ui
+++ b/atef/ui/config_group.ui
@@ -135,13 +135,6 @@
          </layout>
         </item>
         <item>
-         <widget class="QPushButton" name="add_devices_button">
-          <property name="text">
-           <string>+ Add Device</string>
-          </property>
-         </widget>
-        </item>
-        <item>
          <spacer name="verticalSpacer">
           <property name="orientation">
            <enum>Qt::Vertical</enum>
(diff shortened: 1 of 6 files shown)


klauer:
This is in pretty good shape, I think.

Is there any way we can get access to the device list from the `IdAndCompWidget`/bridge in order to pre-fill the search widget again?

ZLLentz:
> Is there any way we can get access to the device list from the IdAndCompWidget/bridge in order to pre-fill the search widget again?

It's a totally different dataclass object, so I think we need to add an argument to `IdAndCompWidget` to accept the `QDataclassList` with the device names in it. And then we need to add that into the `widget_args` for the `AtefItem` that sets up the `IdAndCompWidget`

ZLLentz:
Works as advertised :+1:
Not sure what's up with the sizing, the names on the right get absolutely destroyed in the real estate department. This might be fine or it might matter once I add the navigation buttons.

ZLLentz:
Aside: I need to refactor `AtefItem` now that I've given up on lazy loading. The existing setup is just confusing.

klauer:
I got sidetracked with attempting to do threaded loading of happi metadata to show on normal device selection. Even with the threading, it adds a ~0.5 second delay every time you select a different item, annoyingly enough. I'll stash this work in a separate branch for later review.

Considering we're running out of prep time for Thursday, I'd say if this worked for you, let's merge as-is and revisit later.

PR #71, opened 2022-05-03
REF/ENH: navigation improvements

ZLLentz:
## Description

Original Scope:
- [x] Clickable up arrow on every display to go up the tree to parent items
- [x] Clickable down or right or something arrow to go down the tree to child items
- [x] Restore keyboard navigation to the tree widget

Other tweaks:
- Remove extra margins from the two-column sections to give them a little bit more room.
- Remove or hide spacers below device/signal selection widgets to make these widgets take up the whole space instead of half of it as the ui is expanded.
- Refactor the tree navigation components a bit to make it easier to consider the navigation separately from the widgets. This includes the introduction/standardization of `PageWidget` and `link_page`.
- Add a no filename clause at `DesignerDisplay` so I could create a base class that inherits from `DesignerDisplay` without loading a file. This allowed me to step around an issue where `DesignerDisplay` needs to be the last thing before `QWidget` in the `mro` usually so that all the widgets can be created before we start setting them up.
- Minor adjustments to the requirements file to make the micromamba ci work

## Motivation and Context

It should be easy to find the fields you need to edit.
Previously, the only way to browse through the config was to click around in the tree widget. Now, you can use the arrow keys in the tree widget and you can also click buttons next to subitems to check what settings they have defined.

## How Has This Been Tested?

Interactively but needs more testing with loading files, etc.
I actually don't have happi set up on my local computer which needs to be resolved so I can give the whole thing a real run-through.

I did some docstrings as I went, but I do need to rerun through this page and make sure they all make sense- it would have been easy to mess something up here with the refactor.

image

### DIFF
--- a/atef/ui/comp_view.ui
+++ b/atef/ui/comp_view.ui
@@ -20,6 +20,13 @@
       <number>0</number>
      </property>
      <item>
+      <widget class="QToolButton" name="parent_button">
+       <property name="text">
+        <string>...</string>
+       </property>
+      </widget>
+     </item>
+     <item>
       <widget class="QLabel" name="label">
        <property name="text">
         <string>Name:</string>
@@ -44,13 +51,6 @@
         </size>
        </property>
       </spacer>
-     </item>
-     <item>
-      <widget class="QToolButton" name="parent_button">
-       <property name="text">
-        <string>...</string>
-       </property>
-      </widget>
      </item>
     </layout>
    </item>
--- a/atef/ui/comp_view.ui
+++ b/atef/ui/comp_view.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>400</width>
-    <height>358</height>
+    <height>364</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -44,6 +44,13 @@
         </size>
        </property>
       </spacer>
+     </item>
+     <item>
+      <widget class="QToolButton" name="parent_button">
+       <property name="text">
+        <string>...</string>
+       </property>
+      </widget>
      </item>
     </layout>
    </item>
--- a/atef/ui/config_group.ui
+++ b/atef/ui/config_group.ui
@@ -150,19 +150,6 @@
           </property>
          </layout>
         </item>
-        <item>
-         <spacer name="verticalSpacer">
-          <property name="orientation">
-           <enum>Qt::Vertical</enum>
-          </property>
-          <property name="sizeHint" stdset="0">
-           <size>
-            <width>20</width>
-            <height>40</height>
-           </size>
-          </property>
-         </spacer>
-        </item>
        </layout>
       </widget>
      </item>
--- a/atef/ui/config_group.ui
+++ b/atef/ui/config_group.ui
@@ -121,8 +121,17 @@
      <item>
       <widget class="QWidget" name="devices_container" native="true">
        <layout class="QVBoxLayout" name="devices_container_layout">
+        <property name="leftMargin">
+         <number>0</number>
+        </property>
+        <property name="topMargin">
+         <number>0</number>
+        </property>
+        <property name="rightMargin">
+         <number>0</number>
+        </property>
         <property name="bottomMargin">
-         <number>1</number>
+         <number>0</number>
         </property>
         <item>
          <widget class="QLabel" name="label_4">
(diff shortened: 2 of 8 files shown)


ZLLentz:
Tree/widget references are a big confusing.
1. Every tree item needs a reference to a widget for us to show when the tree item is selected.
2. All widgets that show directly from the tree need a reference to their own tree item to able to create more sub-items. They need this reference at init time (or, we could just do this later maybe?) so that they can set up signals/slots appropriately (or, do they not need this at init time? We could redirect the slots to self methods instead of directly to the tree item reference?)
3. All of these widgets also need a reference to the top-level tree and their item's parent item to be able to direct the tree to select the parent when we click the "go to parent" up button.
4. The "go to child" buttons need references to the child tree items and to the top-level tree to be able to direct the tree to select the child when we click the "go to child" down button

After typing this all out, I think the following refactor is needed:
1. Widgets need to be instantiable without a tree item
2. Tree items should be created after a widget is created
3. Tree items should be assigned a widget before being added to the tree
4. When a widget is assigned to a tree item, it should be given appropriate references to the item, the parent item, and to the whole tree
5. When a widget creates a sub-display, it needs to retain a reference to the sub-display's tree item

ZLLentz:
Yeah, this totally got sidetracked...

ZLLentz:
I've hit a good progress point here. Some refactoring has been done, all the existing pages have navigation buttons, and keyboard navigation works in the tree widget. I'm going to read through the diff and document any other tweaks I made that weren't in the initial scope and take some screenshots before marking as ready for review.

ZLLentz:
Warning: I need to test this further with some real configs, but I'm not intending to change the core here

klauer:
Poking around with this, mostly in place of an in-depth review.

Mostly pertinent to this PR's changes:
* Love the new navigation buttons. I think a tooltip saying "Navigate to (parent name)" or "Navigate to comparison xyz" could be a nice addition.
* Since the name can be of varying length, should the navigation button be to the left of the name textbox? This would make them align nicely
* Similarly, I think the "Navigate to parent" may be better off on the left of the item name? Just an opinion; UX might be a bit better
* Was slightly confused when loading an existing filename from the CLI didn't work - apparently the branch is behind a few commits. I probably would have remembered that a month ago...

Other thoughts/hopes:
* Some default sizing issues (worthy of an issue, probably); do we need more splitters perhaps? Screenshot below
<details>
image
</details>

* Device/component selector: it'd be nice if double-click worked for selection of device names and component names. I thought we had made it this way but it isn't working for components for me.
* Comparison listing: a summary in a tooltip or a widget to the side might be nice
* I like the tab group that lists different files, but I wonder if it might get in the way of multitasking/making side-by-side comparisons of configurations. That said, you could always just open two `atef` instances. So not really an issue.

I think that's about it for now. This makes the GUI so much nicer to use 👏

ZLLentz:
This is probably a few commits behind, my bad- I should definitely merge or rebase. I think all your positioning suggestions are valid and I will adjust appropriately and add tooltips.

Sizing issues didn't happen on my local gui. I bet some of these buttons/widgets need larger minimum sizes... I need to stare at that screenshot a bit and compare with my local view and see what the settings are.

Some of the other things I can make issues for.

ZLLentz:
I moved/updated the buttons and it does feel nicer now
I need to figure out what I broke with respect to file loading- opening the basic demo file crashes now, CLI or otherwise

klauer:
> edit: I need to adjust the setup/loading of some of these widgets because they now need to be able to initialize independently of the tree widget existing and being available

Is this related to the `AttributeError: 'Overview' object has no attribute 'full_tree'` exception I'm seeing when trying the last commit?

ZLLentz:
Yes, this is exactly that- none of the widgets get a reference to the tree on init any more because I wanted to disentangle the widgets and the tree. I made some small tweaks to cause the pages to set up on getting a tree assigned instead of immediately on init.

Works correctly now (loads the demo config):
image

klauer:
Confirmed the latest commit fixed it on my end, too 👍 

One more sizing problem for an issue - the window is too large and I can't see all the PVs on my small laptop monitor 😁 
<details>

image

</details>

ZLLentz:
That's not just a you problem- I think the PV entry widget needs a full rework since it's likely to be used like this.
https://github.com/pcdshub/atef/issues/74
https://github.com/pcdshub/atef/issues/75

I think what I want to do is as described in these two issues- add a scrollbar and switch to a multi-line text box that we splitline on save rather than using the same ui mechanisms as the configuration tags.

ZLLentz:
Ok I'm going to add a few issues based on your comments and merge this, and then I want to adjust the PV widget in a follow-up PR because I think that's the final major usability issue here. There's so much more to improve but I think that is a good breakpoint for switching to the startup speed saga.

klauer (approved):
Code LGTM and functions as promised. Will move other discussion out to issues instead of continuing it here.

PR #72, opened 2022-05-04
ENH: show metadata in device selection + improve performance

klauer:
## Description
* Show happi metadata when selecting devices
* Reuses the Item search/metadata/Component widget and just hides the component selector
    * In principle we could keep it visible, it would just slow things down unnecessarily since we wouldn't accept attribute selections there...

## Motivation and Context
* Code mentioned in #70 
* Fixes mistaken `client.__len__` call for a performance increase (`len(client)` implicitly researched the database)
* Re-uses cached metadata instead of hitting the happi client again

## How Has This Been Tested?
Interactively

image

### DIFF
--- a/atef/widgets/config.py
+++ b/atef/widgets/config.py
@@ -36,7 +36,7 @@ from ..qt_helpers import QDataclassBridge, QDataclassList, QDataclassValue
 from ..reduce import ReduceMethod
 from ..type_hints import PrimitiveType
 from .core import DesignerDisplay
-from .happi import HappiDeviceComponentWidget, HappiSearchWidget
+from .happi import HappiDeviceComponentWidget
 from .ophyd import OphydAttributeData, OphydAttributeDataSummary
 
 logger = logging.getLogger(__name__)
@@ -1437,11 +1437,18 @@ class DeviceListWidget(StringListWithDialog):
         to_select : list of str, optional
             If provided, the device chooser will filter for these items.
         """
-        self._search_widget = HappiSearchWidget(client=util.get_happi_client())
-        self._search_widget.happi_items_chosen.connect(self.add_items)
+        self._search_widget = HappiDeviceComponentWidget(
+            client=util.get_happi_client(),
+            show_device_components=False,
+        )
+        self._search_widget.item_search_widget.happi_items_chosen.connect(
+            self.add_items
+        )
         self._search_widget.show()
         self._search_widget.activateWindow()
-        self._search_widget.edit_filter.setText(util.regex_for_devices(to_select))
+        self._search_widget.item_search_widget.edit_filter.setText(
+            util.regex_for_devices(to_select)
+        )
 
 
 class ComponentListWidget(StringListWithDialog):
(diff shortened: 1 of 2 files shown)


klauer:
OK, this is another performance increasing PR mostly, but also switches up the device search a little bit to show metadata.

I'm open to suggestions - if the original search widget was better without metadata, for example.
Overall, though, I think this is better than what's in master.

ZLLentz:
I'm :+1: on showing the metadata here, I'll browse the code and give it a try

ZLLentz:
This is so smooth interactively to the point where happi basically needs this as a built-in `happi browse` command

atef/widgets/config.py
```diff
@@ -1437,11 +1437,18 @@ def _open_device_chooser(self, to_select: Optional[List[str]] = None):
...
+            show_device_components=False,
+        )
+        self._search_widget.item_search_widget.happi_items_chosen.connect(
+            self.add_items
+        )
         self._search_widget.show()
         self._search_widget.activateWindow()
-        self._search_widget.edit_filter.setText(util.regex_for_devices(to_select))
+        self._search_widget.item_search_widget.edit_filter.setText(
+            util.regex_for_devices(to_select)
```
ZLLentz, line 1450: Note to self: functionally what this does for the device selector is allow us to double-click a device to view its metadata

atef/widgets/happi.py
```diff
@@ -345,18 +358,25 @@ def _update_metadata(self):
...
+            try:
+                self.item = self.client[self.item_name]
+            except KeyError:
+                self.item = None
+            return dict(self.item or {})
 
-        metadata = dict(self.item or {})
+        self._worker = ThreadWorker(get_metadata)
+        self._worker.returned.connect(self._got_metadata)
+        self._worker.start()
```
ZLLentz, line 373: Good to do these in threads for anyone who hasn't updated to the latest happi, or in case something else becomes slow or otherwise goes wrong in the future

atef/widgets/happi.py
```diff
@@ -423,27 +453,35 @@ class HappiDeviceComponentWidget(DesignerDisplay, QWidget):
...
+        if not self.show_device_components:
+            self.device_tab_widget.removeTab(0)
+            self.setWindowTitle("Happi Item Search with Metadata")
 
     @QtCore.Slot(list)
     def _new_item_selection(self, items: List[str]) -> None:
         """New item selected from the happi search."""
         client = self.client
-        if not client or not items:
+        if client is None or not items:
```
ZLLentz, line 484: good catch

atef/widgets/happi.py
```diff
@@ -475,7 +512,23 @@ def set_device(device: Optional[ophyd.Device] = None):
         item, *_ = items
 
         # Set metadata early, even if instantiation fails
-        self.metadata_widget.item_name = item
+        # self.metadata_widget.item_name = item
```
ZLLentz, line 515: Dangling comment
```suggestion
```

ZLLentz (approved):
I like this a lot

PR #85, opened 2022-05-05
ENH: demo config

ZLLentz:
## Description

Simple demo config using some of the LFE devices
This is currently just dropped into the top-level folder, it probably should be changed to a demos folder?

## Motivation and Context

For the demo

## How Has This Been Tested?

Interactively, in front of people

N/A


ZLLentz:
Not meant to be a super in-depth PR, just adding a folder to track demo configs to give a more realistic look at what these sorts of setups will look like to a user and how they would behave.

This particular config does need some adjustments to be truly useful outside the contexts of demos/testing.

klauer (approved):
Makes perfect sense to have this in the repo for our demo purposes 👍 

If it were in the package, we could even make sure we don't regress and break loading somehow?

ZLLentz:
Right, I didn't consider any potential merits of putting it in the package. Regression testing is a clear use case. I'm happy to drop it a level to `atef/demo-configs`?

klauer:
There is `atef/tests/configs/*.yml`, but it does seem to be buried unnecessarily deeply for demo configs.
I guess you could just toss it there for now and we revisit it later?

ZLLentz:
I want to be able to find and use this later. Tests can reach out into the package more cleanly than users can reach into the tests. Semantically, something in the `tests` folder I expect to mean "this exists entirely because of testing".

I'm going to leave it exactly where it is with the understanding that we can move it later when we want to use it for regression testing.

PR #88, opened 2022-05-05
FIX: populate starting list

### LINKED ISSUE #77: BUG: device and component selectors initialize blank when loading a file
## Expected Behavior

These widgets should initialize with the devices and components I selected last time

## Current Behavior

They do not

klauer:
## Description
Populate list when opening a config file

## Motivation and Context
Closes #77 

## How Has This Been Tested?
Interactively

image

### DIFF
--- a/atef/widgets/config.py
+++ b/atef/widgets/config.py
@@ -1072,10 +1072,10 @@ class StringListWithDialog(DesignerDisplay, QWidget):
             adding a new dataclass to the list, which means we should
             definitely append it.
         """
-        if not self.allow_duplicates and item in self.data_list.get():
-            return
-
         if not init:
+            if not self.allow_duplicates and item in self.data_list.get():
+                return
+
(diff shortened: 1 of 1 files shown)


ZLLentz:
I'll take a moment to wrap my ahead around why this works I think
In retrospect, this sort of api with the `init` arg is unclear and confusing- with that implementation, there could have been a better arg like `mutate` or something, but even better there should have been two functions- one to add to the gui, and one to add to the gui while extending the underlying data structure, so the could would read more explicitly and we wouldn't be mixing these sorts of things together

klauer:
I think the `init` arg is just fine, mainly so long as it's part of a private method and not a public one.
`init` could be `sync_with_bridge` or `add_to_bridge`

ZLLentz (approved):
This is clear to me now, not sure why I was hung up before :+1:

PR #93, opened 2022-06-10
ENH: use a QListWidget and a multiline dialog to edit PV lists

### LINKED ISSUE #74: Feature: way to add PVs in bulk for PV Configuration
## Expected Behavior

We should be able to copy/paste a bunch of PV config PVs at once.

## Current Behavior

One at a time

## Possible Solution

Multiline text box?

## Context

It takes a long time to build out a PV config

### LINKED ISSUE #75: BUG: PV Configs have no scrollbar
See title
Gets long really fast

ZLLentz:
## Description

Replace the tag-like editing of PVs with a QListWidget that is edited primarily with a multiline text edit.

When you click the + button, a multiline text edit appears. Anything you put into this box is added to the list.
When you select one or more existing lines and double-click, a multiline text edit appears with the lines you selected. You can add, edit, or remove items as you like. On confirmation the new lines will be compared to the old lines and we'll add or remove items as appropriate.

## Motivation and Context

Usability fix- it's very annoying to edit and browse the PV lists now. This lets you copy paste big lists of PVs if you want.

closes #74 
closes #75 

## How Has This Been Tested?

Interactively only

One oddity: the drag/drop here still needs to be implemented to update the data class. I think this probably affects the device/component widgets too, if someone tries to rearrange the list it will have no internal effect and therefore will not be saved.

N/A

image
image
image

### DIFF
--- a/atef/ui/id_and_comp.ui
+++ b/atef/ui/id_and_comp.ui
@@ -92,13 +92,6 @@
         <item>
          <layout class="QVBoxLayout" name="id_content"/>
         </item>
-        <item>
-         <widget class="QPushButton" name="add_id_button">
-          <property name="text">
-           <string>Add Identifier</string>
-          </property>
-         </widget>
-        </item>
        </layout>
       </widget>
      </item>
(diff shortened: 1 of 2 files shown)


ZLLentz:
Remaining annoyances:
- Editing an existing PV is awkward because if you add one after it it gets appended to the very end of the list
- Drag-and-drop doesn't mutate the internal dataclass

After some testing though I think I'll be done here for now

klauer (approved):
Aside from the 2 things you already mentioned, I think this works really well!

PR #96, opened 2022-06-23
ENH: expand root nodes in happi device search tree during search

JJL772:
## Description

Expands the root nodes of the happi device tree if the filter box isn't empty. Restores the previous expansion state of the root nodes once the filter box is emptied out. 

## Motivation and Context

It's a bit more convenient to search the tree if groupings are automatically expanded. Otherwise you need to manually click the arrows to actually see the search results.

## How Has This Been Tested?

Tested this interactively only by running searches and clicking on things.

Shouldn't affect any other areas of the code. 

N/A

Categories are expanded fully during search:
image

And restored to the original state after the filter box is empty:
image

### DIFF
--- a/atef/widgets/happi.py
+++ b/atef/widgets/happi.py
@@ -51,6 +51,7 @@ class HappiSearchWidget(DesignerDisplay, QWidget):
     _client: Optional[happi.client.Client]
     _last_selected: List[str]
     _search_thread: Optional[ThreadWorker]
+    _tree_view_expanded: Optional[List[QtCore.QModelIndex]]
     _tree_current_category: str
     _tree_updated: bool
     button_choose: QtWidgets.QPushButton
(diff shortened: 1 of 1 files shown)


atef/widgets/happi.py
```diff
@@ -51,6 +51,7 @@ class HappiSearchWidget(DesignerDisplay, QWidget):
     _client: Optional[happi.client.Client]
     _last_selected: List[str]
     _search_thread: Optional[ThreadWorker]
+    _tree_view_expanded: Optional[List[QtCore.QModelIndex]]
```
klauer, line 54: Since we're stashing `QModelIndex` with , it probably should be cleared when the happi model updates underneath it, I'd imagine. Probably as simple as setting `_tree_view_expanded` in `refresh_happi`

klauer (approved):
Nice! The tree not expanding bugged me before - so this is a great addition.

Issue #40, opened 2022-04-19
Bug: Config GUI tag deletion crash (QWidget attribute .text)


ZLLentz:
## Expected Behavior

A crash has been reported on tag deletion

## Current Behavior

This may or may not be the right widget for it, but I tried to delete the last tag and crashed with the following:

```python
(atef) PC98125:atef klauer$ atef config
Traceback (most recent call last):
  File "/Users/klauer/Repos/atef/atef/bin/config.py", line 1197, in eventFilter
    if object.text():
AttributeError: 'QWidget' object has no attribute 'text'
```

_Originally posted by @klauer in https://github.com/pcdshub/atef/pull/28#discussion_r853314450_

klauer:
I saw this one recently when redoing the device list widgets.

ZLLentz:
I saw this also today while setting up the demo config

PR #97, opened 2022-06-29
BUG: fix issue with running event filter after cleanup

ZLLentz:
## Description

Check that we are dealing with a line edit in our line edit event filter before proceeding to the checks and style updates.

## Motivation and Context

There is an event filter on some of our edit fields that handles things like changing style when the user selects/deselects the widget. This is only installed on compatible object types, but it can also be called on generic `QWidget` objects that are suspiciously named `line_edit` after the line edit has been removed/cleaned up. You see an unimportant error message when removing some of these objects since `QWidget` doesn't have a `text` attribute, which is checked during this filter. The application continues to run as normal, since the only thing skipped is updating the style of a removed widget, but the error popup is annoying and potentially confusing.

I found this after a report that removing comparisons did not work, which I could not reproduce, but I did find this on the way.

Traceback:
```
  File "C:\Users\zlentz\Miniconda3\envs\dev\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\zlentz\Miniconda3\envs\dev\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\zlentz\Documents\VSCode\atef\atef\__main__.py", line 3, in <module>
    main()
  File "C:\Users\zlentz\Documents\VSCode\atef\atef\bin\main.py", line 94, in main
    func(**kwargs)
  File "C:\Users\zlentz\Documents\VSCode\atef\atef\bin\config.py", line 46, in main
    app.exec()
  File "C:\Users\zlentz\Documents\VSCode\atef\atef\widgets\config.py", line 1869, in eventFilter
    if object.text():
AttributeError: 'QWidget' object has no attribute 'text'
```

## How Has This Been Tested?

Interactively only

It'll go in the release notes, this is a bugfix.
I added a comment since it isn't immediately clear why this clause would be needed.

### DIFF
--- a/atef/widgets/config.py
+++ b/atef/widgets/config.py
@@ -1860,6 +1860,10 @@ class FrameOnEditFilter(QObject):
     not editing it.
     """
     def eventFilter(self, object: QLineEdit, event: QEvent) -> bool:
+        # Even if we install only on line edits, this can be passed a generic
+        # QWidget when we remove and clean up the line edit widget.
+        if not isinstance(object, QLineEdit):
+            return False
         if event.type() == QEvent.FocusIn:
             object.setFrame(True)
             object.setReadOnly(False)


JJL772 (approved):

tangkong (approved):
LGTM.  I appreciate the comment as I slowly get acclimated to atef stuff

klauer (approved):
I believe this closes https://github.com/pcdshub/atef/issues/40

ZLLentz:
Yes, that's correct- closes #40 (though this didn't cause a crash any more)

I'm personally surprised by the underlying pyqt behavior described here, but :shrug:

PR #106, opened 2022-07-11
ENH: add scrollbars to comparison list and checklist list, adjust style

### LINKED ISSUE #103: No Scroll Option for "Add Checklist" Page
## Expected Behavior
When adding checklists, there should be a point where a scroller comes up and allows for more room for checklists.

## Current Behavior
When creating checklists, the window just expands until it fills the whole screen, limiting the amount of checklists a configuration can have.

## Possible Solution
Add a scrolling feature to this page to allow for more checklists.

1. When creating a configuration, click on "Add Checklist"
2. Repeat 1. until the page fills the window.
3. At this point, the user is no longer able to add more configurations.

## Context
This limits the size of each configuration and forces the checks to be split up between multiple configurations.

## Your Environment
source pcds_conda
My repo: /cds/home/l/liggett1/atef

ZLLentz:
## Description

Add scrollbars to the two sections where we can add large numbers of sub-items.

## Motivation and Context

Otherwise, the gui just expands relentlessly as we add items.
closes #103 

## How Has This Been Tested?

Interactively only

Will include in the release notes

From an empty config, here's me jamming the "add checklist" button:
image

A look at the demo config which doesn't overflow the scroll limit of the default size on my screen:
image

### DIFF
--- a/atef/widgets/config.py
+++ b/atef/widgets/config.py
@@ -1874,6 +1874,14 @@ class FrameOnEditFilter(QObject):
 
     @staticmethod
     def set_edit_style(object: QLineEdit):
+        """
+        Set a QLineEdit to the look and feel we want for editing.
+
+        Parameters
+        ----------
+        object : QLineEdit
+            Any line edit widget.
+        """
         object.setFrame(True)
         object.setStyleSheet(
             "QLineEdit { background: white }"
@@ -1882,6 +1890,14 @@ class FrameOnEditFilter(QObject):
 
     @staticmethod
     def set_no_edit_style(object: QLineEdit):
+        """
+        Set a QLineEdit to the look and feel we want for not editing.
+
+        Parameters
+        ----------
+        object : QLineEdit
+            Any line edit widget.
+        """
         if object.text():
             object.setFrame(False)
             object.setStyleSheet(
--- a/atef/widgets/config.py
+++ b/atef/widgets/config.py
@@ -1790,8 +1790,8 @@ class StrListElem(DesignerDisplay, QWidget):
     def __init__(self, start_text: str, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.line_edit.setText(start_text)
-        self.line_edit.setFrame(not start_text)
         edit_filter = FrameOnEditFilter(parent=self)
+        edit_filter.set_no_edit_style(self.line_edit)
         self.line_edit.installEventFilter(edit_filter)
         self.on_text_changed(start_text)
         self.line_edit.textChanged.connect(self.on_text_changed)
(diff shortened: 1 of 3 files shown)


liggett2820:
it would be helpful to put one on this page too
Screen Shot 2022-07-11 at 4 52 01 PM

ZLLentz:
Looks like it carries over to here too- looks ok, space inefficiency aside:
image

atef/widgets/config.py
```diff
@@ -1865,16 +1865,30 @@ def eventFilter(self, object: QLineEdit, event: QEvent) -> bool:
...
-            object.setReadOnly(True)
+            self.set_no_edit_style(object)
             return True
         return False
 
+    @staticmethod
+    def set_edit_style(object: QLineEdit):
+        object.setFrame(True)
+        object.setStyleSheet(
+            "QLineEdit { background: white }"
```
klauer, line 1879: (I think this is good as-is, just wondering aloud...)

I wonder if this would be possible to do in a single QSS block with 2 state-based styles (https://doc.qt.io/qt-5/stylesheet-reference.html; active vs non-active or something). 

The read-only part definitely wouldn't be, but the frame/background may be.

atef/widgets/config.py
```diff
@@ -1865,16 +1865,30 @@ def eventFilter(self, object: QLineEdit, event: QEvent) -> bool:
...
         if event.type() == QEvent.FocusOut:
-            if object.text():
-                object.setFrame(False)
-            object.setReadOnly(True)
+            self.set_no_edit_style(object)
             return True
         return False
 
+    @staticmethod
+    def set_edit_style(object: QLineEdit):
```
klauer, line 1876: Docstrings would be nice if you have a chance

klauer (approved):
Tried it out and it works as advertised. Looks good to me 👍

atef/widgets/config.py
ZLLentz: I'll add some docstrings

atef/widgets/config.py
ZLLentz: So, instead of using a custom filter on focus events for the background, the alternative here is to use a pseudo-state selector to set the background style? e.g., "in this state, background transparent" (in other states, no change).
For example, perhaps `QLineEdit[:no_frame] { background: transparent }`

ZLLentz: I tried for a bit to get this to work but I can't figure out the precise invocation and stylesheet debug is pain. Skipping (but still editing to use the proper palette color)

tangkong (approved):
This looks good to me, I keep putting off really diving into ATEF but I will by the next review request I promise.

PR #107, opened 2022-07-12
ENH: Restore correct background color when locking/unlocking edits

JJL772:
## Description

When switching between locked & unlocked states for the overview name/description, the background color changes to white to indicate that the widget is editable. On dark themes this makes it unreadable.

## Motivation and Context

On dark themes, the text would become unreadable when switching between locked and edit mode for the overviews.

## How Has This Been Tested?

Tested by clicking lock/unlock

n/a

## Screenshots

Unlocked editing. Before this change, the background color of the QLineEdits would be pure white. 
image

Locked
image

### DIFF
--- a/atef/widgets/config.py
+++ b/atef/widgets/config.py
@@ -905,7 +905,7 @@ class OverviewRow(ConfigTextMixin, DesignerDisplay, QWidget):
             color = self.palette().color(QPalette.ColorRole.Base)
             self.setStyleSheet(
                 f"QLineEdit, QPlainTextEdit {{ background: rgba({color.red()},"
-                f"{color.green()}, {color.blue()}) }}"
+                f"{color.green()}, {color.blue()}, {color.alpha()}) }}"
             )
             if not self.name_edit.text():
                 self.delete_button.setEnabled(True)
--- a/atef/widgets/config.py
+++ b/atef/widgets/config.py
@@ -902,10 +902,10 @@ class OverviewRow(ConfigTextMixin, DesignerDisplay, QWidget):
             self.delete_button.setEnabled(False)
         else:
             self.desc_edit.setFrameShape(self.desc_edit.StyledPanel)
-            color = self.palette().color(QPalette.ColorRole.AlternateBase)
+            color = self.palette().color(QPalette.ColorRole.Base)
             self.setStyleSheet(
                 f"QLineEdit, QPlainTextEdit {{ background: rgba({color.red()},"
-                f"{color.green()}, {color.blue()}, 255) }}"
+                f"{color.green()}, {color.blue()}) }}"
             )
             if not self.name_edit.text():
                 self.delete_button.setEnabled(True)
--- a/atef/widgets/config.py
+++ b/atef/widgets/config.py
@@ -18,7 +18,7 @@ from pydm.widgets.drawing import PyDMDrawingLine
 from qtpy import QtWidgets
 from qtpy.QtCore import QEvent, QObject, QTimer
 from qtpy.QtCore import Signal as QSignal
-from qtpy.QtGui import QColor
+from qtpy.QtGui import QColor, QPalette
 from qtpy.QtWidgets import (QAction, QCheckBox, QComboBox, QFileDialog,
                             QFormLayout, QHBoxLayout, QInputDialog, QLabel,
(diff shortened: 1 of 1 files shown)


atef/widgets/config.py
```diff
@@ -902,8 +902,10 @@ def handle_locking(self, locked: bool):
             self.delete_button.setEnabled(False)
         else:
             self.desc_edit.setFrameShape(self.desc_edit.StyledPanel)
+            color = self.palette().color(QPalette.ColorRole.AlternateBase)
             self.setStyleSheet(
-                "QLineEdit, QPlainTextEdit { background: white }"
+                f"QLineEdit, QPlainTextEdit {{ background: rgba({color.red()},"
+                f"{color.green()}, {color.blue()}, 255) }}"
```
klauer, line 908: Ignoring the alpha channel seems reasonable

atef/widgets/config.py
```diff
@@ -902,8 +902,10 @@ def handle_locking(self, locked: bool):
             self.delete_button.setEnabled(False)
         else:
             self.desc_edit.setFrameShape(self.desc_edit.StyledPanel)
+            color = self.palette().color(QPalette.ColorRole.AlternateBase)
```
klauer, line 905: [Qt docs](https://doc.qt.io/qt-5/qpalette.html#ColorRole-enum) indicate:

Constant | Value | Description
-- | -- | --
QPalette::Base | 9 | Used  mostly as the background color for text entry widgets, but can also be  used for other painting - such as the background of combobox drop down  lists and toolbar handles. It is usually white or another light color.
QPalette::AlternateBase | 16 | Used as the alternate background color in views with alternating row colors (see QAbstractItemView::setAlternatingRowColors()).

Hmm, what does `Base` look like with your theme? 

`Base` looks correct on macOS:
image
image

`AlternateBase` on the other hand is a bit duller:
image
image

A separate bug appears to be that the on-load stylesheet is different in either case - or not applied somehow:
image

JJL772: image
Base looks good to me!

klauer (approved):
LGTM - code and visually 👍

ZLLentz:
Good catch- I'll also review (and then use this palette thing in my other PR which is hard-coding a color to white..)

atef/widgets/config.py
```diff
@@ -902,8 +902,10 @@ def handle_locking(self, locked: bool):
             self.delete_button.setEnabled(False)
         else:
             self.desc_edit.setFrameShape(self.desc_edit.StyledPanel)
+            color = self.palette().color(QPalette.ColorRole.Base)
             self.setStyleSheet(
-                "QLineEdit, QPlainTextEdit { background: white }"
+                f"QLineEdit, QPlainTextEdit {{ background: rgba({color.red()},"
```
ZLLentz, line 907: It took me a while to find all the options for color specification since I was curious where the rgba thing came from: https://doc.qt.io/qt-5/stylesheet-reference.html#list-of-property-types

```
rgb(r, g, b)
| rgba(r, g, b, a)
| hsv(h, s, v)
| hsva(h, s, v, a)
| hsl(h, s, l)
| hsla(h, s, l, a)
| #rrggbb
| [Color Name](https://doc.qt.io/qt-5/qcolor.html#setNamedColor)
```

Should we include the `color.alpha()` channel? It seems like a arbitrary omission

klauer: In the first proposal it was just set to `255`, then removed, and now set to the palette's alpha value as in 744cc49. I think the latter is the most correct, so let's go with it.

atef/widgets/config.py
```diff
@@ -902,8 +902,10 @@ def handle_locking(self, locked: bool):
...
+                f"{color.green()}, {color.blue()}) }}"
```
klauer, line 908:
```suggestion
                f"{color.green()}, {color.blue()}, {color.alpha()}) }}"
```

atef/widgets/config.py
JJL772: Didn't see this comment until now- my bad!

atef/widgets/config.py
ZLLentz: I'm going to pull this suggestion in and merge

Issue #100, opened 2022-07-06
Comparisons to Dynamic Values


ZLLentz:
## Expected Behavior

It should be possible to add comparisons to PV values and signal values rather than just to static values.

## Current Behavior

All comparisons are currently to static values.

## Possible Solution

One way to handle this is to allow any of the value fields to hold `str` types that hold references to other data.
For example, we could borrow the pydm channel syntax and interpret something like `"ca://MY:PVNAME"` as "use channel access to get MY:PVNAME and use the value as the comparison." We could refer to signals as `"sig://my_signame"` or use other more creative addressing as appropriate.

## Context

Andrew has been working on some laser system automated check setups and raised to our attention that this is an important feature.

ZLLentz:
After mulling it over at the Alpine Inn, I'm fully on board. Sitting in the `pydm` box is going to be annoying (and, truthfully, maybe `pydm` should consider extending the channel definition itself instead of filling up increasingly elaborate strings).

There must be a clean implementation for this via creating a `DynamicValue` dataclass parallel to `Value`, either completely separate or inheriting from `Value`. There are some wrinkles with the proliferation of this to things like the > comparisons or the range comparisons which don't use `Value` at all right now but need to start doing so.

tangkong:
Looking into resurrecting this effort, before moving on to other Active checkout steps and features.  We should refer back to #111 as work continues, but the GUI has moved considerably since its last commit (Jul 2022), and we'll have to do more than merge the diff.

tangkong:
closed by #184

PR #111, opened 2022-07-14
WIP: dynamic value loading

ZLLentz:
## Description

First look at how we might implement comparing against dynamic values
Config gui is broken here
Needs to be revisited and finished

## Motivation and Context

Make PR to not lose track of this effort
Related to #100 

## How Has This Been Tested?

WIP

WIP

### DIFF
--- a/atef/check.py
+++ b/atef/check.py
@@ -263,6 +263,9 @@ class Comparison:
     #: result severity.
     if_disconnected: Severity = Severity.error
 
+    def __post_init__(self):
+        self.is_prepared: bool = False
+
     def __call__(self, value: Any) -> Optional[Result]:
         """Run the comparison against ``value``."""
         return self.compare(value)
@@ -306,6 +309,10 @@ class Comparison:
             An identifier that goes along with the provided value.  Used for
             severity result descriptions.
         """
+        if not self.is_prepared:
+            raise UnpreparedComparisonException(
+                f"Comparison {self} was not prepared."
+            )
         if value is None:
             return Result(
                 severity=self.if_disconnected,
@@ -395,19 +402,30 @@ class Comparison:
             )
 
     def prepare(self) -> None:
-        """Implement in subclass to grab and cache dynamic values."""
+        """
+        Implement in subclass to grab and cache dynamic values.
+
+        This is expected to set self.is_prepared to True if
+        successful.
+        """
         raise NotImplementedError()
 
 
 @dataclass
-class Equals(Comparison):
-    value: PrimitiveType = 0.0
+class BasicDynamic(Comparison):
     value_dynamic: Optional[DynamicValue] = None
+
+    def prepare(self) -> None:
+        if self.value_dynamic is not None:
+            self.value = self.value_dynamic.get()
+        self.is_prepared = True
+
+
+@dataclass
+class Equals(BasicDynamic):
+    value: PrimitiveType = 0.0
     rtol: Optional[Number] = None
     atol: Optional[Number] = None
-
-    def __post_init__(self):
-        self.is_prepared = False
 
     @property
     def _value(self) -> Value:
@@ -428,29 +446,15 @@ class Equals(Comparison):
         return f"{comparison}{dynamic}{self._value}"
 
     def _compare(self, value: PrimitiveType) -> bool:
-        if not self.is_prepared and self.value_dynamic is not None:
-            raise UnpreparedComparisonException(
-                f"Dynamic equals comparison to {self.value_dynamic} "
-                "was not prepared."
-            )
         return self._value.compare(value)
-
-    def prepare(self) -> None:
-        if self.value_dynamic is not None:
-            self.value = self.value_dynamic.get()
-        self.is_prepared = True
 
 
 @dataclass
-class NotEquals(Comparison):
+class NotEquals(BasicDynamic):
     # Less confusing shortcut for `Equals(..., invert=True)`
     value: PrimitiveType = 0
-    value_dynamic: Optional[DynamicValue] = None
     rtol: Optional[Number] = None
     atol: Optional[Number] = None
-
-    def __post_init__(self):
-        self.is_prepared = False
 
     @property
     def _value(self) -> Value:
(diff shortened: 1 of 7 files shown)


ZLLentz:
Rebased on top of master- resuming work on this

ZLLentz:
I'm not entirely sold on this approach but I can't immediately think of anything better. The idea here is: anywhere that was previously a `PrimitiveType` or a `Number` is now a `Union` with that and `DynamicValue`. `DynamicValue` classes have multiple flavors and all the comparisons know that a `DynamicValue` class has a `get` method that must be called at runtime to get the current value. `DynamicValue` subclasses each contain metadata that helps us understand how to get the value.

In the file, this will show up as a serialized `EpicsValue` or `HappiValue` dataclass dictionary in cases where we have a dynamic value.

In the GUI, we will need to make a special value input widget where the user can right click -> change mode. The modes will be e.g. "bool" mode, "int" mode, all the static modes, and then "epics" mode and "happi" mode for the available dynamic values.

klauer:
> I don't know if I should wrap with a StaticValue class to keep things uniform or if it's OK as-is.

* It's certainly serialized more cleanly when left as-is.
* API-wise, it'd be cleaner not to have to do `isinstance()`-then-maybe-`.get()` were `StaticValue` to be wrapped

Some other thoughts:
* If we were to add `StaticValue` with the standard serializer/deserializer, it'd break existing configuration files (but we could customize (de)serialization to avoid that, I _think_)
* You should try to use the `atef.cache` mechanism to avoid re-creating the same signal over and over again. `DynamicValue` will have to be able to access that cache instance somehow and the best way doesn't seem immediately obvious. This is also important for the test suite, as it has a stand-in cache that doesn't require PV connections.
* I wonder if this may need tighter integration with `PreparedComparison` such that the preparation step would do the `DynamicValue` setup, caching, and comparison running (https://github.com/klauer/atef/blob/d08da1a8d40a1f40d44402372b7bdb9f26222fe4/atef/config.py#L655)

> In the GUI, we will need to make a special value input widget where the user can right click -> change mode. 

I like the idea of being able to switch easily between the modes - this sounds good.

ZLLentz:
All those points are valid.

I'll leave as-is for now RE: no `StaticValue`.
I'll look at `atef.cache` before proceeding (currently I'm doing some local caching).
After that, I'll  try to get something working with the GUI (including value preview?).
At the end I'll look at optimizing the `PreparedComparison` integration.

klauer:
More unsolicited thoughts (sorry):
* `Comparison.compare()` _probably_ shouldn't call `.get()`; or if it does, it needs to be come async-friendly (which isn't so ophyd-compatible just yet)

I'm thinking keeping the comparisons free of everything but the basic comparisons (e.g., no connection logic) could be a good thing. 

To that end, what if:
* `Comparison(value=EpicsValue(pvname="abc"))` remains valid for configuration files + serialization purposes
*  `Comparison(value=EpicsValue(pvname="abc")).compare(0)` raises e.g. `UnpreparedComparisonException` because it doesn't support dynamic value comparisons
*  The "prepare" step could perform a conversion from dynamic to static:
     * We can plug in the caching/connection/etc layer more dynamically
     * `convert_to_static_comparison(Comparison(value=EpicsValue(pvname="abc")), cache=signal_cache)`
     * The above converts it to, for example, `Comparison(value=5.0)` 
     * If there are any errors in doing that conversion, the CLI/etc will be able to report more precisely where it went wrong

ZLLentz:
v0 ui file of the entry widget. I intend to connect all the widgets and show only the one selected by the user or deduced from the load. From top to bottom: bool, enum, float, epics, happi, int, str (alphabetic only to make it easier to manage in the designer ui)
image

ZLLentz:
I think you're right about the "preparation" step needing to "static-ify" a dynamic comparison. I had some vague realization of this following your previous comment but hadn't put much thought into the specifics and instead thought about the UI.

The other way to approach this I think is to have a `prepare` routine that caches the current static value. Then instead of the `Union`/`hasattr`/`get` jungle, we just add an `dynamic_value = Optional[DynamicValue]` parameter to the various data classes and call `prepare` as appropriate: a routine that does nothing special for `dynamic_value = None` but replaces the value if we do have a dynamic value configured. Calling `compare` would still know if we'd ever prepared or not and raise as appropriate.

We could still use `get` or `prepare` outside the context of running a test suite if needed, e.g. for showing preview values in the config gui.

ZLLentz:
I think I will opt for as described in my previous message for the purpose of cleaning up the isinstance jungle. If at the end we decide that replacing the full dataclass is better I think this is still cleaner in that implementation.

One neat small thing this does is allow us to save a static value in the serialized file and restore it if the user changes their mind. But this isn't a decision consideration.

ZLLentz:
Ui is looking decent, still has a few kinks to work out though:
- Right click menu to adjust turned out to be awkward because a lot of widgets (like line edits) have native right click menus
- Range widget still needs to be converted to use the new entry widget
- Some of the displays get really wide when you select a long pv or device name
- Unclear if the current representations are ideal
- int spinbox has left-aligned text and a fixed width
- No controls for e.g. a relative PV or signal attr

The enum finder somehow worked on the first try and everything else was a bit uphill.

ZLLentz:
I'm currently reviewing the other PR and I realized that I may need to consider reduction settings for dynamic comparison targets

ZLLentz:
Some screenshots of the ui edits:

New comparison, default setup. Mouseover the new button to see the mode.
image

Click the button to switch modes.
image

Enum mode picks up the enum strings of your configured PVs/signals.
image

EPICS and Happi are the dynamic modes. EPICS has a text-entry field for the PV, happi pops open the component dialog. The refresh button collects a sample value for the UI.
image

(This is too wide and some sizing/layout adjustments are needed, but otherwise I like it)

ZLLentz:
Maybe we should add an archive mode here too?

klauer:
Another couple interesting ones could be (for far down the line...):
* "Value from another comparison" (taking care for circular references...)
* "Value from another configuration file"

ZLLentz:
I'll revisit this after making the GUI catch up to the checking code

tangkong:
superceded by #184 , which heavily borrowed from this work.

PR #114, opened 2022-07-14
ENH: "tools" such as ping in passive checkout procedures

### LINKED ISSUE #113: Feature: Add "tools" such as "ping" that can be run in passive checkouts
For starters:
* Ping (parameters: hosts)

Others:
* Check IOCs via IOC manager (parameters: ioc names)

Per the discussion in https://github.com/pcdshub/atef/discussions/99#discussioncomment-3059844

klauer:
## Description
* In addition to PV configurations and Device configurations, this PR will add Tool configurations.
* A preliminary "ping" tool is included here.
* Test suite has been updated to test tools
* pytest-asyncio has been added as a test dependency for coroutines in the test suite
* Tests which previously performed comparisons have been adjusted to work with this
* Starting to move checks to be asyncio-friendly, where possible
* Added `DataCache` caching mechanism for suites of passive checks / comparisons to be run with
* Added opt-in parallel mode for `atef check` (`atef check --parallel/-p`) which aggressively acquires all data for comparisons prior to running

## Motivation and Context
Closes #113 

## TODO
* Additional tools will follow based on discussions.
* GUI updates will come separately.

## How Has This Been Tested?
Test suite and simple external test config:

<details>

```json
{
  "configs": [
    {
      "ToolConfiguration": {
        "name": "Host alive check",
        "description": "Tool",
        "tags": null,
        "tool": {
          "Ping": {
            "hosts": ["1.1.1.1", "8.8.8.8"]
          }
        },
        "checklist": [
          {
            "name": "Check that the ping time is good",
            "ids": [
              "max_time"
            ],
            "comparisons": [
              {
                "Less": {
                  "name": "Ping time OK",
                  "description": "Is the ping less than x ms?",
                  "invert": false,
                  "reduce_period": null,
                  "reduce_method": "average",
                  "string": null,
                  "severity_on_failure": 2,
                  "if_disconnected": 2,
                  "value": 5
                }
              }
            ]
          }
        ]
      }
    }]
}

```
</details>

Issues, discussions, and this PR text so far

```
❌ Error Host alive check (Tool)
└── Check that the ping time is good
    └── Ping time OK: Is the ping less than x ms?
        └── ❌ Error: max_time < 5: value of 19.669666666666668
```

### DIFF
--- a/atef/bin/check.py
+++ b/atef/bin/check.py
@@ -3,6 +3,7 @@
 """
 
 import argparse
+import asyncio
 import dataclasses
 import logging
 import pathlib
@@ -14,6 +15,7 @@ import rich
 import rich.console
 import rich.tree
 
+from ..cache import DataCache, _SignalCache, get_signal_cache
 from ..check import Comparison, Result, Severity
 from ..config import (AnyConfiguration, Configuration, ConfigurationFile,
                       PathItem, PreparedComparison)
@@ -50,6 +52,12 @@ def build_arg_parser(argparser=None):
         nargs="*",
         dest="name_filter",
         help="Limit checkout to the named device(s) or identifiers",
+    )
+
+    argparser.add_argument(
+        "-p", "--parallel",
+        action="store_true",
+        help="Acquire data for comparisons in parallel",
     )
 
     return argparser
@@ -262,12 +270,49 @@ async def check_and_log(
     verbose: int = 0,
     client: Optional[happi.Client] = None,
     name_filter: Optional[Sequence[str]] = None,
+    parallel: bool = True,
+    cache: Optional[DataCache] = None,
 ):
-    """Check a configuration and log the results."""
+    """
+    Check a configuration and log the results.
+
+    Parameters
+    ----------
+    config : AnyConfiguration
+        The configuration to check.
+    console : rich.console.Console
+        The rich console to write output to.
+    verbose : int, optional
+        The verbosity level for the output.
+    client : happi.Client, optional
+        The happi client, if available.
+    name_filter : Sequence[str], optional
+        A filter for names.
+    parallel : bool, optional
+        Pre-fill cache in parallel when possible.
+    cache : DataCache
+        The data cache instance.
+    """
     items = []
     name_filter = list(name_filter or [])
     severities = []
-    for prepared in PreparedComparison.from_config(config, client=client):
+
+    if cache is None:
+        cache = DataCache()
+
+    all_prepared = list(
+        PreparedComparison.from_config(config, client=client, cache=cache)
+    )
+
+    cache_fill_tasks = []
+    if parallel:
+        for prepared in all_prepared:
+            if isinstance(prepared, PreparedComparison):
+                cache_fill_tasks.append(
+                    asyncio.create_task(prepared.get_data_async())
+                )
+
+    for prepared in all_prepared:
         if isinstance(prepared, PreparedComparison):
             if name_filter:
                 device_name = getattr(prepared.device, "name", None)
@@ -316,8 +361,10 @@ async def main(
     filename: str,
     name_filter: Optional[Sequence[str]] = None,
     verbose: int = 0,
+    parallel: bool = False,
     *,
-    cleanup: bool = True
+    cleanup: bool = True,
+    signal_cache: Optional[_SignalCache] = None,
 ):
     path = pathlib.Path(filename)
     if path.suffix.lower() == ".json":
@@ -333,6 +380,7 @@ async def main(
         client = None
 
     console = rich.console.Console()
+    cache = DataCache(signals=signal_cache or get_signal_cache())
     try:
         with console.status("[bold green] Performing checks..."):
             for config in config_file.configs:
@@ -342,6 +390,8 @@ async def main(
                     verbose=verbose,
                     client=client,
                     name_filter=name_filter,
+                    parallel=parallel,
+                    cache=cache,
                 )
     finally:
         if cleanup:
--- a/atef/bin/check.py
+++ b/atef/bin/check.py
@@ -256,7 +256,7 @@ def log_results_rich(
     console.print(root)
 
 
-def check_and_log(
+async def check_and_log(
     config: AnyConfiguration,
     console: rich.console.Console,
     verbose: int = 0,
@@ -285,7 +285,7 @@ def check_and_log(
                     )
                     continue
 
-            prepared.result = prepared.compare()
+            prepared.result = await prepared.compare()
             if prepared.result is not None:
                 items.append(prepared)
                 severities.append(prepared.result.severity)
@@ -312,7 +312,7 @@ def check_and_log(
     )
 
 
-def main(
+async def main(
     filename: str,
     name_filter: Optional[Sequence[str]] = None,
     verbose: int = 0,
@@ -336,7 +336,7 @@ def main(
     try:
         with console.status("[bold green] Performing checks..."):
             for config in config_file.configs:
-                check_and_log(
+                await check_and_log(
                     config,
                     console=console,
                     verbose=verbose,
--- a/atef/cache.py
+++ b/atef/cache.py
@@ -6,7 +6,7 @@ import dataclasses
 import logging
 import typing
 from dataclasses import dataclass, field
-from typing import (Any, Dict, Hashable, Mapping, Optional, Sequence, Type,
+from typing import (Any, Dict, Hashable, Iterable, Mapping, Optional, Type,
                     TypeVar, cast)
 
 import ophyd
@@ -87,14 +87,27 @@ class ToolKey:
 
 
 def _freeze(data):
-    if isinstance(data, (int, float, str)):
+    """
+    Freeze ``data`` such that it can be used as a hashable key.
+
+    Parameters
+    ----------
+    data : Any
+        The data to be frozen.
+
+    Returns
+    -------
+    Any
+        Hopefully hashable version of ``data``.
+    """
+    if isinstance(data, str):
         return data
     if isinstance(data, Mapping):
         return frozenset(
             (_freeze(key), _freeze(value))
             for key, value in data.items()
         )
-    if isinstance(data, Sequence):
+    if isinstance(data, Iterable):
         return tuple(_freeze(part) for part in data)
     return data
 
--- a/atef/cache.py
+++ b/atef/cache.py
@@ -1,26 +1,51 @@
 from __future__ import annotations
 
+import asyncio
+import concurrent.futures
+import dataclasses
 import logging
+import typing
 from dataclasses import dataclass, field
-from typing import Dict, Mapping, TypeVar
+from typing import (Any, Dict, Hashable, Mapping, Optional, Sequence, Type,
+                    TypeVar, cast)
 
 import ophyd
 
-_CacheSignalType = TypeVar("_CacheSignalType")
+from .reduce import ReduceMethod, get_data_for_signal_async
+from .type_hints import Number
+
+if typing.TYPE_CHECKING:
+    from . import tools
+
+_CacheSignalType = TypeVar("_CacheSignalType", bound=ophyd.Signal)
 
 
 logger = logging.getLogger(__name__)
+_signal_cache: Optional[_SignalCache[ophyd.EpicsSignalRO]] = None
+
+
+def get_signal_cache() -> _SignalCache[ophyd.EpicsSignalRO]:
+    """Get the global EpicsSignal cache."""
+    global _signal_cache
+    if _signal_cache is None:
+        _signal_cache = _SignalCache[ophyd.EpicsSignalRO](ophyd.EpicsSignalRO)
+    return _signal_cache
 
 
 @dataclass
 class _SignalCache(Mapping[str, _CacheSignalType]):
-    signal_type_cls: _CacheSignalType
+    signal_type_cls: Type[ophyd.EpicsSignalRO]
     pv_to_signal: Dict[str, _CacheSignalType] = field(default_factory=dict)
 
     def __getitem__(self, pv: str) -> _CacheSignalType:
         """Get a PV from the cache."""
         if pv not in self.pv_to_signal:
-            self.pv_to_signal[pv] = self.signal_type_cls(pv, name=pv)
+            signal = cast(
+                _CacheSignalType,
+                self.signal_type_cls(pv, name=pv)
+            )
+            self.pv_to_signal[pv] = signal
+            return signal
 
         return self.pv_to_signal[pv]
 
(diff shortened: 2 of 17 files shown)


atef/bin/check.py
```diff
@@ -312,7 +312,7 @@ def check_and_log(
     )
 
 
-def main(
+async def main(
```
ZLLentz, line 315: Intuitively this seems like a really good design decision to me as we add various tools that may take some time to do their checks in non-uniform ways. Does this open the possibility of running checks in parallel and recombining their results later?

klauer: I think so, too.  Moving to asyncio should make running things in parallel easier among other things.

atef/tools.py
```diff
@@ -0,0 +1,200 @@
...
+
+from . import serialization
+from .check import Result, Severity
+
+#: Arguments that can be passed to tools.
+ToolArguments = Dict[str, Any]
+T = TypeVar("T", bound="Tool")
+
+
+class ToolResult(TypedDict):
```
klauer, line 20: My logic for switching to `TypedDict` here instead of a `dataclass` for each tool was along the lines of data interchange:
* External tools could more easily return JSON blobs that get deserialized into Python dictionaries
* Maybe such tools could also optionally take JSON as input via stdin or less preferably argv

Thinking a bit further - there's no reason we couldn't use `apischema.(de)serialize` here to take a dataclass and serialize it to JSON easily.

I think the argument for `TypedDict` is a bit weak here, so I'd be OK switching this back.

atef/tools.py
```diff
@@ -0,0 +1,200 @@
...
+        host : str
+            The host to ping.
+
+        Returns
+        -------
+        PingResult
+        """
+
+        if sys.platform == "win32":
+            args = ("/n", str(self.count))
```
klauer, line 133: Untested on Windows

atef/tools.py
```diff
@@ -0,0 +1,200 @@
...
+    The base result dictionary of any tool.
+    """
+    result: Result
+
+
+class PingResult(ToolResult):
+    """
+    The result dictionary of the 'ping' tool.
+    """
+    #: Host(s) that are alive
```
klauer, line 31: For the results of `ping`:
* `min_time` and `max_time` are probably the most useful and also the easiest to use with some `Comparison` (`<= 500ms` for example)
* Values that return a list (as in `alive` or `unresponsive`) or a dictionary (as in `times`) are probably not easy to use with `Comparison`
   * Does that indicate that the tool definition is poor?
   * Does it indicate that new comparisons may be required?

I haven't really decided yet, but I'm throwing this out there

klauer: I mitigated this sub-keys being valid in the identifier list (e.g., `hosts.hostname` could be valid).
The splitting method can still use some work as we wouldn't want `hosts.google.com` to turn into `("hosts", "google", "com")`.

atef/config.py
```diff
@@ -144,18 +175,111 @@ class PreparedComparison:
...
+
+        Raises
+        ------
+        TimeoutError
+            If unable to connect or retrieve data from the signal.
+        """
+        if self.signal is None:
+            raise ValueError("Signal instance unset")
+        # TODO: async?
+        self.data = self.comparison.get_data_for_signal(self.signal)
```
ZLLentz, line 279: I wonder if it's possible or desirable to build some async paths into the ophyd `Signal` classes for these sorts of applications.
```
class Signal(...):
    async def get_async(self, **kwargs):
        return some_future_suspiciously_similar_to_a_status
```
(if it's not super clear I haven't worked much with asyncio directly)

atef/config.py
```diff
@@ -179,8 +304,37 @@ def compare(self) -> Result:
...
+                reason=(
+                    f"Getting data for signal {self.identifier!r} comparison "
+                    f"{self.comparison} raised {ex.__class__.__name__}: {ex}"
+                ),
+            )
+
+        if self.data is None:
+            # 'None' is likely incompatible with our comparisons and should
+            # be raised for separately
+            # TODO reviewers: Is this a sensible assumption?
```
ZLLentz, line 327: I cannot think of any situation where the user is checking a control system value and is expecting the Python `None` as a good result. Pyepics sometimes drops `None` in for disconnections, right?

atef/config.py
```diff
@@ -193,7 +347,7 @@ def from_device(
         comparison: Comparison,
         name: Optional[str] = None,
         path: Optional[List[PathItem]] = None,
-    ) -> PreparedComparison:
+    ) -> PreparedSignalComparison:
```
ZLLentz, line 350: Slightly off-topic: are these `from_device` and `from_pvname` methods the proper place to get and stash the dynamic values in #111? (and `from_tool` for the off chance someone wants to... compare a ping ms to a PV?)

atef/config.py
```diff
@@ -179,8 +304,37 @@ def compare(self) -> Result:
                     f"for comparison {self.comparison}"
                 ),
             )
-        self.result = self.comparison.compare_signal(
-            self.signal,
+
+        try:
+            self.data = await self.get_data()
```
ZLLentz, line 309: I'm surprised that this bit and the corresponding tool runs aren't done when we prepare the comparison. Maybe I misunderstand the staging here about what is expected in the prepare step vs the compare step. I'm slightly concerned about this mainly for the dynamic configurations where we might get each side of the comparison inequality at different points in time.

atef/tools.py
ZLLentz: I think it's a bit strange that this is the only TypedDict in the sea of atef dataclasses. But it's hard for me to have a bigger opinion about which is better to use since they are extremely similar result-wise.

atef/tools.py
ZLLentz: It may be useful to summarize difficult to compare keys e.g. `num_alive`, `num_unresponsive`. If we want to support comparisons of sequence types, then maybe we could also include reduction options like `len` that drop them down to a scalar more generically?

atef/tools.py
ZLLentz: My windows machine has nothing useful to ping except maybe I can make a config that pings google

atef/tools.py
```diff
@@ -0,0 +1,295 @@
...
+
+        ping_by_host: Dict[str, Union[Exception, PingResult]] = {}
+
+        async def _ping(host: str) -> None:
+            try:
+                ping_by_host[host] = await self.ping(host)
+            except Exception as ex:
+                ping_by_host[host] = ex
+
+        tasks = [asyncio.create_task(_ping(host)) for host in self.hosts]
```
ZLLentz, line 271: Is this line and the implications of doing this in general the reason why you async-ified the checking process?

ZLLentz (commented):
Lots of thoughts and questions, not very many suggestions

atef/tools.py
klauer: `localhost` will work just fine (see the test suite + included config file)

atef/config.py
klauer: There might be something we can do here, but it'll take a bit of investigation.

1. Things that can happen asynchronously - and report asynchronous status via callbacks can be relatively easily turned asyncio-compatible with some cross-communication between a callback and an asyncio Task
2. Things that have busy loops, such as the [poor mechanism pyepics uses](https://github.com/pyepics/pyepics/blob/762b96fe10dfb0194e3a5ae3eb9df8fe0a04b2c1/epics/ca.py#L1523-L1529) in the background (and ophyd inherits necessarily) 

For the pyepics example in (2), the hack would have to happen at roughly the `epics.ca` or `epics.PV` level, making it not so compelling/desirable to modify

atef/config.py
klauer: Yes, pyepics returns `None` which [ophyd interprets](https://github.com/bluesky/ophyd/blob/384d1a96c2a8956b5c3b3df62c7d72379485ba5a/ophyd/signal.py#L1216-L1220) as `TimeoutError` - so we shouldn't actually see it.

I think I'll remove this comment and leave in the check - unless we ever find a scenario where `None` could be meaningful in atef.

atef/config.py
klauer: I'd say that's pretty on-topic as your PR's needs should drive the API choices in here as well.

As of now, "prepare" means "take comparison, device, attribute and create the outer product of possible checks".

Given that we want to grab external values once, I'd say those dynamic values probably _should_ be retrieved here (just once, no matter how many times they are used).

It'd be nice to not block on `from_device` while making up all of these prepared comparisons, though. I think this needs some more thought.

atef/config.py
klauer: I think you're right and this needs some work, similar to the above.

atef/tools.py
klauer: I think that task parallelization (be it for the tools, comparisons, or anything else in atef) will be important for design decisions and of course performance.  I realized that adding threads would be going in the direction of the "old" ophyd and would create trouble down the line for us (with v2 being dangled in front of us).

It's not too late to switch back to threaded, if you feel that it's premature to go the asyncio route.

ZLLentz: The only problem with the async approach is the one we outlined above: ophyd/pyepics do not (cannot) support it easily, e.g. any aync ophyd method is probably just a threaded method in disguise

atef/tools.py
ZLLentz: Would it be helpful then to simply run the test suite on my windows machine?

atef/config.py
ZLLentz: > It'd be nice to not block on from_device while making up all of these prepared comparisons, though. I think this needs some more thought.

It sounds like this needs some sort of async-light callback that stashes the value when it comes in and flags that the values is ready

atef/tools.py
klauer: Let's think about both scenarios a bit - my reply above was not well-thought out enough. Enough so to probably be misleading.

Any big things I might be missing here?

### Threading
* We want to parallelize general tasks such as the tools provided in this PR, so we spawn threads and wait on some completion `threading.Event` or utilize a `ThreadPool`
   1. These are ugly and tougher to maintain/understand, but they work
   2. Propagation of exceptions/errors gets a bit more complicated, but it's doable
* We want `ophyd.EpicsSignal.get()` and other blocking ophyd stuff to be done in parallel
   1. As in the general task parallelization, we similarly put such calls into background threads
   2. `.wait_for_connection()` and `.get()` occasionally yield control due to their usage of `time.sleep()`, at which point other threads can run
* Data reduction/subscription handling is done via callback with the acquisition thread blocking (no changes really necessary here)

### asyncio
* Task parallelization as in this PR is done rather tersely with `create_task`/`wait`
* We want to somehow take non-async ophyd stuff and integrate it with asyncio
   1. We pass `.get()` calls off to run in the asyncio event loop's [executor thread pool](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor)
   2. Assuming ophyd's own threading doesn't get in the way, it should be as simple as `value = await loop.run_in_executor(None, signal.get)`
   4. `.wait_for_connection()` and `.get()` will occasionally yield control due to `time.sleep()`, at which point other threads (including our asyncio Event loop thread) can run
* We'll want to move subscription/data reduction processing to asyncio (looks like I already [started that](https://github.com/klauer/atef/blob/c2d65ce16a21f6d5463f3682baad8c1800540ee7/atef/ophyd_helpers.py#L217))

##### My take

I think that even though the `.get()` replacement is "a threaded method in disguise", if we want to be asyncio-friendly in our development, it'd probably be easier to build it in now rather than later. It will certainly not be perfect, but I expect it could run well enough for what we need - and the code will probably be cleaner.

But you shouldn't just take my word on this; I'll dig a bit deeper into the above and we can revisit.

atef/tools.py
klauer: If you wouldn't mind, that could help. I doubt we'll ever get a CI job for Windows configured.
Alternatively, if you want to run `atef check -vv atef/tests/configs/ping_localhost.json` that'll do, too.

atef/tools.py
klauer: > I think it's a bit strange that this is the only TypedDict in the sea of atef dataclasses. 

Your comment plus my lack of good reason for it above is enough to say it was a mistake to go this route.  I'm switching it back to a dataclass. 

The `get_result_value_by_key` is flexible enough that it works with mappings/objects, so the conversion is very easy now.

atef/tools.py
ZLLentz: Mixed results here: the test suite fails and the atef check passes.
```
_____________________________________________________ test_result_severity[all_good] ______________________________________________________ 

conf = ToolConfiguration(name=None, description=None, tags=None, checklist=[IdentifierAndComparison(name=None, ids=['max_time...27.0.0.1'], 
times={'127.0.0.1': 100.0}, min_time=0.0, max_time=100.0), hosts=['127.0.0.1'], count=1, encoding='utf-8'))
severity = <Severity.success: 0>

    @config_and_severity
    @pytest.mark.asyncio
    async def test_result_severity(
        conf: ToolConfiguration, severity: Severity
    ):
        overall, results = await config.check_tool(conf.tool, conf.checklist)
>       assert overall == severity
E       assert <Severity.error: 2> == <Severity.success: 0>

atef\tests\test_comparison_tools.py:61: AssertionError
```

klauer: Hmm, running something like `pytest --pdb --lf`, what's in the `results` list?

atef/tools.py
ZLLentz: I need to brush up on how asyncio works. I didn't know about `run_in_executor` for example. I didn't necessarily mean to apply that a "threaded method in disguise" is necessarily a very bad thing- it just seems to be the only reason why we wouldn't go the async route.

klauer: Implications of the question aside - I think this is a really good thing to discuss. (Whether a line comment in a PR is the right spot for it or not, meh...)

That is to say:
* We definitely don't want to flippantly decide on big architectural things like this
* I should be able to solidly justify why I think we should just go asyncio now rather than later
* The implications of threaded/asyncio should be clear before we commit

ZLLentz: I'm at the same point as before where it intuitively seems right to me to go to async to help the checker run at scale, but I would have trouble verbalizing an argument for it.

I think we're going from single threaded to async here and not doing a threaded implementation at all which seems more correct than doing it the other way to me. The main advantage of async over threaded is brevity and stricter control over the execution compared to letting the python thread scheduler take the wheel, right? Theoretically fewer race conditions?

atef/tools.py
ZLLentz:
```
[Result(severity=<Severity.error: 2>, reason='max_time <= 1: value of 100.0')]
```
Does windows ping... have a different time scale?

ZLLentz: I'm a bit confused but I'll look a little more closely
```
C:\Users\zlentz>ping localhost

Pinging PC98346.win.slac.stanford.edu [127.0.0.1] with 32 bytes of data:
Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
Reply from 127.0.0.1: bytes=32 time<1ms TTL=128

Ping statistics for 127.0.0.1:
    Packets: Sent = 4, Received = 4, Lost = 0 (0% loss),
Approximate round trip times in milli-seconds:
    Minimum = 0ms, Maximum = 0ms, Average = 0ms
```

klauer: Odd, I think the regular expression didn't match any times in `ping` in that case. The 100 is the default value to use on error (separate bug that reminded me of: need to normalize times in seconds instead of the awkward `ms`)

ZLLentz: I think the problem is that the output from windows is formatted completely differently

klauer: Oh, did it by chance switch units to microseconds/nanoseconds? I thought I recalled a platform that did that, but neither macOS nor Linux were deviating from `ms`

klauer: I'll switch up the regular expression to match `<1ms` as just `1ms`. I hope that'll be good enough.

atef/tools.py
klauer: Added `num_*` variants. I think we can probably push back the decision on the reduction options for other tools.

A _better_ simple check here could just be `num_unresponsive == 0`

klauer:
With the discussion above, I've come to realize that `PreparedComparison` still needs quite a bit of refactoring.

It should rather be per-device, per-PV, or per-tool: and should support one or more checks on the data retrieved from those disparate sources. It needs to be broken up at the level that's sensible for avoiding repeating operations unnecessarily (running the same tool multiple times to try a different comparison, acquiring the same data from the same device signal multiple times, etc.) and parallelization.

I'm not sure the above is coherent; I'm still rethinking the approach. Hoping to meet in the middle with the current API and not too much reworking. More on this in a bit.

klauer:
I had implemented what I thought was the right approach (per my previous comment) and then realized it was too much of a mess. Instead, I've implemented a data caching layer that can be reused for any/all comparisons in a set.

I believe this is a better solution overall:
* `PreparedComparison` remains a single comparison coupled with a tool/PV/signal
* Data caching is external to the `PreparedComparison` so the caller can determine where it's appropriate to pass a different `DataCache` instance (or clear it, etc)
* It's a lot easier to pass the cached information around, inspect it, and manage it
* asyncio tasks make managing the cache a bit easier than expected

Other stuff:
* Opt-in for parallel mode `p` with `atef check` means any ophyd/device related bugs would hopefully not be showstoppers.
* Future improvement: don't unnecessarily create tasks in the cache initialization step in parallel mode

atef/tools.py
klauer: I think the parallel data cache implementation shows that this can be cleaner and viable even now with ophyd as-is. I'm reasonably convinced and hope @ZLLentz will be, too

(More on this later, it's lunchtime)

atef/check.py
```diff
@@ -258,6 +259,11 @@ def compare(self, value: Any, identifier: Optional[str] = None) -> Result:
         if self.invert:
             passed = not passed
 
+        # Some comparisons may be done with array values; require that
+        # all match for a success here:
+        if isinstance(passed, Iterable):
```
klauer, line 264: I ran into this failure locally where `passed` was `np.ndarray([True, False, False])` because I had messed around with the caproto simple IOC used in a sample config. If you try `if np.ndarray([...])` you'll find that it raises and helpfully notes "use `all()`!"

Comparisons against numpy arrays needs to make its way into the test suite itself.

klauer (commented):
I realize there's a lot here and this will be tough to properly and fully review.

I think if we agree on the overall approach and there's not much in the way of code smell, we can probably proceed with this and work on figuring out how to integrate dynamic values into it.

atef/check.py
klauer: On the other hand, this could be the wrong approach.

What about if we detect iterable values prior to running the comparison and reduce them via the configured data reduction method? For example, instead of blindly running `compare([1, 2, 3])` we reduce that to `compare((1 + 2 + 3) / 3)`

atef/config.py
```diff
@@ -140,22 +172,162 @@ class PreparedComparison:
...
+        - Including data reduction settings
+    """
+    #: The device the comparison applies to, if applicable.
+    device: Optional[ophyd.Device] = None
+    #: The signal the comparison is to be run on.
+    signal: Optional[ophyd.Signal] = None
+    #: The value from the signal the comparison is to be run on.
+    data: Optional[Any] = None
+
+    async def get_data_async(self) -> Any:
```
ZLLentz, line 301: Note to self: this is where we'd put the bit that gets the dynamic comparison's values, if any exist.

ZLLentz (approved):
I'm very :+1: on this formulation. I had minor questions and comments only. I didn't respond to the other thread during this review because I didn't want to mess up the discussion page again but I'll comment after.

atef/check.py
ZLLentz: I think an all match is a reasonable stand-in, but I think probably we'd prefer to move to deliberate reductions (where `all` could be a reduction, somehow?). I might be intending to compare against an array but I also might be trying to target a scalar PV and getting the array by mistake.

atef/bin/check.py
```diff
@@ -256,18 +264,55 @@ def log_results_rich(
...
-    for prepared in PreparedComparison.from_config(config, client=client):
+
+    if cache is None:
+        cache = DataCache()
+
+    all_prepared = list(
+        PreparedComparison.from_config(config, client=client, cache=cache)
+    )
+
+    cache_fill_tasks = []
```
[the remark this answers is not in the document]
klauer: It's only here because I read the docs for once:
https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
> Important
> 
> Save a reference to the result of this function, to avoid a task disappearing mid execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage-collected at any time, even before it’s done. For reliable “fire-and-forget” background tasks, gather them in a collection:

atef/cache.py
```diff
@@ -1,26 +1,51 @@
...
+from typing import (Any, Dict, Hashable, Mapping, Optional, Sequence, Type,
+                    TypeVar, cast)
 
 import ophyd
 
-_CacheSignalType = TypeVar("_CacheSignalType")
+from .reduce import ReduceMethod, get_data_for_signal_async
+from .type_hints import Number
+
+if typing.TYPE_CHECKING:
```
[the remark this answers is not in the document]
klauer: It's particularly helpful for circular imports - the type checker doesn't care about circular imports, but CPython sure does.

atef/cache.py
```diff
@@ -40,12 +65,263 @@ def clear(self) -> None:
...
+    @classmethod
+    def from_tool(cls, tool: tools.Tool) -> ToolKey:
+        settings = cast(Hashable, _freeze(dataclasses.asdict(tool)))
+        return cls(
+            tool_cls=type(tool),
+            settings=settings,
+        )
+
+
+def _freeze(data):
```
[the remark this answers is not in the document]
klauer: I totally agree. 

On this topic - are there any obvious omissions in this function? I put it together only keeping in mind what was in `Ping`

atef/bin/check.py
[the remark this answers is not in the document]
ZLLentz: I thought it might be something like this... I'm surprised that this is enough, but the interaction between scoping/gc isn't always intuitive to me.

atef/check.py
```diff
@@ -103,11 +104,12 @@ def __str__(self) -> str:
     def compare(self, value: PrimitiveType) -> bool:
         """Compare the provided value with this one, using tolerance settings."""
         if self.rtol is not None or self.atol is not None:
-            return np.isclose(
+            close = np.isclose(
                 value, self.value,
                 rtol=(self.rtol or 0.0),
                 atol=(self.atol or 0.0)
             )
+            return close
```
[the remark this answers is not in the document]
klauer: I had the `Iterable` check (topic of the other thread) here and moved it out to `compare()`.  I should revert this one...

atef/check.py
klauer: Very good points. I think you're right - `all` should be a reduction option as well. Probably the default one, for that matter.

If you're OK with it, let's leave this minor change in as a temporary fix and revisit in another PR after further discussion.

atef/bin/check.py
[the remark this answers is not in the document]
klauer: GC definitely can be confusing.

In this specific instance, the list holds a hard reference and persists throughout the coroutine, so it - and the tasks it holds - should only be gc'd after it returns (or raises)

atef/cache.py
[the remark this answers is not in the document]
ZLLentz: No obvious omissions, but the resulting structure isn't 1:1 interface-compatible with the original for mapping types.

Also, I only just realized now, but the first isinstance check for primitives is seemingly redundant unless you're trying to short-circuit two isinstance calls for primitives.

atef/check.py
ZLLentz: Absolutely agree on leaving as-is for this PR while we don't have any built-in options for handling array data

atef/cache.py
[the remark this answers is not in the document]
klauer: In d6d4e04796e92f817bae67e70f51a5df899d255c:
* I tweaked `Sequence` -> `Iterable` (`set()` is `Iterable` but not a `Sequence`, and `list`/`tuple` are all `Iterable` and `Sequence`)
* I took out the short-circuiting of basic types but only left in `str` because it is iterable (and we don't want to freeze each character)

atef/check.py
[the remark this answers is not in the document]
klauer: Cleaned in 7599488bf02ebeb1d61d51d82271fc556b315f88

atef/cache.py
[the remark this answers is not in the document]
ZLLentz: Right, I forgot about str being an iterable- that always bites me randomly

ZLLentz:
I'm happy to merge this and move on to adjusting the dynamic pr on top of it as appropriate

klauer:
~Friday at 4:30 PM? :shipit:~

Yeah, I think I'm done with this for now. Onward to the next big one...

PR #117, opened 2022-07-15
ENH: use row selection mode in Ophyd and Happi table views

JJL772:
## Description

Use row selection mode in the happi and ophyd table views instead of the individual table cell mode.

## Motivation and Context

Row selection mode in these table views makes more sense since the data is related to the rest of the row. 90% of the motivation here is personal preference though, @klauer indicated that row selection mode seemed better too, so I'm not alone in this regard.

## How Has This Been Tested?

Tested by clicking around the table views.

Selection by rows!!!
image


klauer (approved):
LGTM 👍 

Also, just in case you weren't aware, the `pre-commit` settings in the repo can automatically take care of the `isort` for you on each commit.

ZLLentz:
Late :+1: from me

PR #118, opened 2022-07-15
ENH: Context menus for Happi, Ophyd and string list widgets

JJL772:
## Description

This PR adds various context menus to atef. The Happi, Ophyd and string list widgets all got extended context menus. 

For the `HappiSearchWidget`, I added an "Add device" entry to the context menu that emits the `happi_items_chosen` signal. In atef, that means you can right click a device and "Add device" instead of double clicking it to add it to the device list.

For `OphydDeviceTableView`, a context menu helper was added, which is used to create the base `QMenu` that the standard table view actions are added to. `OphydDeviceTableWidget` adds a context menu helper to display the same menu that's displayed when you click the `Select` button at the bottom.

`StringListWithDialog` got a simple context menu that has "Copy" and "Remove" actions. 

## Motivation and Context

Context menus are just more intuitive and quicker to work with than the discrete add/remove/select `QPushButtons`.

## How Has This Been Tested?

Tested by clicking around in the UI.

Some of these might be hard to read because of my system theme :) 

Ophyd signal selection
image

Happi widget
image

String list
image

### DIFF
--- a/atef/widgets/config.py
+++ b/atef/widgets/config.py
@@ -16,8 +16,8 @@ from typing import (Any, Callable, ClassVar, Dict, List, Optional, Tuple, Type,
 from apischema import deserialize, serialize
 from pydm.widgets.drawing import PyDMDrawingLine
 from qtpy import QtWidgets
-from qtpy.QtCore import QEvent, QObject, Qt, QTimer
-from qtpy.QtCore import Signal as QSignal, QPoint
+from qtpy.QtCore import QEvent, QObject, QPoint, Qt, QTimer
+from qtpy.QtCore import Signal as QSignal
 from qtpy.QtGui import QClipboard, QColor, QGuiApplication, QPalette
 from qtpy.QtWidgets import (QAction, QCheckBox, QComboBox, QFileDialog,
                             QFormLayout, QHBoxLayout, QInputDialog, QLabel,
--- a/atef/widgets/config.py
+++ b/atef/widgets/config.py
@@ -1448,9 +1448,9 @@ class StringListWithDialog(DesignerDisplay, QWidget):
 
         def copy_selected():
             items = self.list_strings.selectedItems()
-            if len(items) > 0:
-                QGuiApplication.clipboard().setText(items[0].text(),
-                                                    QClipboard.Mode.Clipboard)
+            text = '\n'.join([x.text() for x in items])
+            if len(text) > 0:
+                QGuiApplication.clipboard().setText(text, QClipboard.Mode.Clipboard)
 
         copy = menu.addAction('&Copy')
         copy.triggered.connect(copy_selected)
--- a/atef/widgets/config.py
+++ b/atef/widgets/config.py
@@ -17,7 +17,7 @@ from apischema import deserialize, serialize
 from pydm.widgets.drawing import PyDMDrawingLine
 from qtpy import QtWidgets
 from qtpy.QtCore import QEvent, QObject, Qt, QTimer
-from qtpy.QtCore import Signal as QSignal
+from qtpy.QtCore import Signal as QSignal, QPoint
 from qtpy.QtGui import QClipboard, QColor, QGuiApplication, QPalette
 from qtpy.QtWidgets import (QAction, QCheckBox, QComboBox, QFileDialog,
                             QFormLayout, QHBoxLayout, QInputDialog, QLabel,
@@ -1431,7 +1431,7 @@ class StringListWithDialog(DesignerDisplay, QWidget):
             else:
                 self._edit_item(old, new)
 
-    def _show_context_menu(self, pos):
+    def _show_context_menu(self, pos: QPoint):
         """
         Displays a context menu that provides copy & remove actions
         to the user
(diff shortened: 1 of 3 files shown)


atef/widgets/config.py
```diff
@@ -1428,6 +1431,32 @@ def edit_items(self, old_items: List[str], new_items: List[str]) -> None:
...
+                QGuiApplication.clipboard().setText(items[0].text(),
+                                                    QClipboard.Mode.Clipboard)
+
+        copy = menu.addAction('&Copy')
+        copy.triggered.connect(copy_selected)
+
+        remove = menu.addAction('&Remove')
+        remove.triggered.connect(self._remove_item_request)
+
+        menu.exec(self.mapToGlobal(pos))
```
klauer, line 1458: I think if there are no selected items, the context menu shouldn't be even shown (as they don't do anything)

atef/widgets/config.py
```diff
@@ -1428,6 +1431,32 @@ def edit_items(self, old_items: List[str], new_items: List[str]) -> None:
             else:
                 self._edit_item(old, new)
 
+    def _show_context_menu(self, pos):
```
klauer, line 1434:
```suggestion
    def _show_context_menu(self, pos: QtCore.QPoint) -> None:
```

atef/widgets/config.py
```diff
@@ -1428,6 +1431,32 @@ def edit_items(self, old_items: List[str], new_items: List[str]) -> None:
...
+        ----------
+        pos : QPoint
+            Position to display the menu at
+        """
+        menu = QMenu(self)
+
+        def copy_selected():
+            items = self.list_strings.selectedItems()
+            if len(items) > 0:
+                QGuiApplication.clipboard().setText(items[0].text(),
```
klauer, line 1449: Multi-selection is possible, but we only copy the first item here?

atef/widgets/happi.py
```diff
@@ -201,7 +201,17 @@ def _list_view_context_menu(self, pos: QtCore.QPoint) -> None:
         """Context menu for the happi list view."""
         self.menu = QtWidgets.QMenu(self)
         index: QtCore.QModelIndex = self.happi_list_view.indexAt(pos)
+
         if index is not None:
+            # Add action to add the selected device
+            add_action = self.menu.addAction(f'&Add Device: {index.data()}')
```
klauer, line 207: I know you're just adding to this menu, perhaps it's worth fixing this bug while you're at it:

If I don't select a device, I still get the menu:
image

atef/widgets/config.py
ZLLentz, line 1434: Perhaps we should annotate this with `pos: QPoint` (from `QtCore`)

ZLLentz (commented):
I think the code looks good here on first read- I'll try the branch next

klauer (commented):
Looking good! Our tastes in user interface interactions are pretty similar 👍

atef/widgets/ophyd.py
```diff
@@ -611,17 +615,24 @@ def copy(*_):
...
-                copy_action = self.menu.addAction(f"&Copy: {index.data()}")
+                copy_action = QtWidgets.QAction(f'&Copy: {index.data()}')
+                standard_actions.append(copy_action)
                 copy_action.triggered.connect(copy)
 
             row_data = self.get_data_from_proxy_index(index)
             if row_data is not None:
-                select_action = self.menu.addAction(
+                select_action = QtWidgets.QAction(
                     f"&Select attribute: {row_data.attr}"
```
ZLLentz, line 628: Can somebody remind me what the function of this "Select attribiute" action was supposed to do? It doesn't seem to do anything on master or here on this branch.

ZLLentz (approved):
This worked as advertised for me. Added one open question about a seemingly pre-existing source of confusion for me.

atef/widgets/ophyd.py
klauer: It should select the attribute and add it to the atef config attribute list. It's not so obvious because it happens in the background - assuming it is still working...

ZLLentz: It doesn't seem to do this any more. I need to click the "select" button at the bottom or one of the new actions Jeremy added here to add to the config.

atef/widgets/config.py
JJL772: I wasn't sure what the user would expect. Right click -> Copy on multiple items only copies the right clicked item, or all of them? Either works for me.

klauer: Personally I'd expect it would copy all of the ones I've selected, perhaps with a newline delimiter. Anyone else feel differently?

JJL772:
This PR will probably need to be squashed

JJL772:
travis is not having a good time right now...
image

klauer:
Ugh, I've restarted Travis manually a few times for other PRs already. I'll poke that one as well.

atef/widgets/ophyd.py
klauer: Confirmed that it doesn't actually work. Looks like the slot wasn't connected somewhere along the way.

ZLLentz:
In some sense, when I open up a failing travis and I see `apt-get install failed`- I can't even be that upset, because my code didn't break the build this time

JJL772:
I tried restarting it myself like 2 hours ago and it still failed! good to know that it listens to Ken at least.

PR #125, opened 2022-08-02
BUG: Json extension bug fix

### LINKED ISSUE #122: New files created in atef config do not save as .json files
## Expected Behavior
All files created in atef config should be saved as .json files.

## Current Behavior
Files created and saved in atef config do not save as .json files.

## Possible Solution
Have files created and saved in atef config save as .json files.

1. Open atef config
2. create new file
3. save file
4. file doesn't save as .json file

## Context
atef won't find the file you saved before when searching for the file to open in atef config

## Your Environment
source pcds_conda
my repo: /cds/home/l/liggett1/atef/

liggett2820:
## Description
New files created in config GUI can now be saved as .json files.

Replaced `with open(filename, 'w') as fd:` with `with open(filename+'.json', 'w') as fd:` on line 189 in [atef/widgets/config.py](https://github.com/pcdshub/atef/blob/a7c875f7f31c81a00eb174e43113866f05c682ec/atef/widgets/config.py#L104)

## Motivation and Context
Closes #122 

## How Has This Been Tested?
Created a test file in the config GUI.

Comment in code.

When opening the GUI, I get these screens.

Screen Shot 2022-08-02 at 10 34 47 AM

Screen Shot 2022-08-02 at 10 34 51 AM

I click on New and come to this screen 

Screen Shot 2022-08-02 at 10 35 06 AM

I then go to File -> Save as... to get to here.

Screen Shot 2022-08-02 at 10 35 23 AM

It then saves as that file.
Screen Shot 2022-08-02 at 10 35 33 AM

I go back to File -> Open and the file appears!

Screen Shot 2022-08-02 at 10 35 40 AM


atef/widgets/config.py
```diff
@@ -186,7 +186,7 @@ def save_as(self, *args, filename: Optional[str] = None, **kwargs):
                 filter='Json Files (*.json)',
             )
         try:
-            with open(filename, 'w') as fd:
+            with open(filename+'.json', 'w') as fd:
```
klauer, line 189: What happens if I type in `my_filename.json` for the filename? Will it become `my_filename.json.json`?

demo-configs/test.json
```diff
@@ -0,0 +1,3 @@
+{
```
klauer, line 1: Please remove these two extra files from your branch (`git rm demo-configs/test.json test_bash_script` then commit and push again)

atef/widgets/config.py
liggett2820: Yes it will. But the goal of this bug fix was so that atef would add the extension automatically in order to be consistent with the "Files of Type" so the user wouldn't need to. atef will still be able to open your file as a .json file.

klauer: I was hoping to prompt you into realizing that the behavior I described is undesirable, but let me put it more succinctly:

* If `.json` is already the suffix of the filename, `.json` should not be added. Multiple filename extensions that have no meaning are at best undesirable or confusing.
* If `.json` is not the suffix of the filename, it should be concatenated to the end of the filename string

If you try a 'save' dialog in something like MS Paint (Word, etc) you'll find that it performs as I described above.

Consider looking at `pathlib`:

```python
In [1]: import pathlib

In [2]: filename = pathlib.Path("/path/to/temp")

In [3]: filename
Out[3]: PosixPath('/path/to/temp')

In [4]: filename.suffix
Out[4]: ''

In [5]: filename = pathlib.Path("/path/to/temp.json")

In [6]: filename.suffix
Out[6]: '.json'
```

liggett2820: I changed it so that I added a check to see if `filename` ends in `.json` instead of always adding the `.json` extension.

atef/widgets/config.py
```diff
@@ -185,6 +185,8 @@ def save_as(self, *args, filename: Optional[str] = None, **kwargs):
                 caption='Save as',
                 filter='Json Files (*.json)',
             )
+        if filename.endswith('.json') is False:   # json_extension_bug_fix location
```
ZLLentz, line 188: Seems reasonable to me. We might still have some weird cases with someone intentionally giving themselves a different extension and getting surprised here, but considering the gui is looking for .json files it's probably OK to force it.

Let's clean up just a little bit: we can remove the comment (the PR/commits are enough documentation) and we don't need to check if something `is False`. The normal python way to do the check is simply:
```suggestion
        if not filename.endswith('.json'):
```

ZLLentz (approved):
Let's merge :+1:
Thanks for finding the bug and for the fix

PR #126, opened 2022-08-02
REF: atef.config rewrite to support arbitrarily grouped hierarchies

### LINKED ISSUE #120: Multiple checklists for device configs
## Expected Behavior
Should be able to assign different checklists to different devices within a device config.

## Current Behavior
One checklist per device config only goes through that one checklist for all of the devices on the device config.

Currently each device config assumes that all devices are the same type. When trying to group devices by MODS TILEs, the checklists don't support different device types.

## Possible Solution
Allow multiple checklists per device config and be able to assign them to different devices.

Screen Shot 2022-07-20 at 2 01 44 PM

## Context
Need to be able to check different devices within each TILE.

## Your Environment
source pcds_conda
my repo: /cds/home/l/liggett1/atef

### LINKED ISSUE #121: Need of new "group config" to incorperate MODS
## Expected Behavior
Below is a rough sketch of the layout of the TILEs in a MODS and how the structure of the json file should look:

MODS

## Current Behavior
Right now, each MODS has to be its own input directory, each TILE has to be its own .json file, and each individual device per each TILE has to have its own config (in the picture below, each individual dot would be a device config).

Screen Shot 2022-07-25 at 11 43 49 AM

## Possible Solution
Add a new config category called "group config". In an ideal world, this config category would allow:
- each MODS have their own .json file
- each TILE would be their own group config
- all of the devices would be able to be listed within the group config
- when adding a device, need an add button on the window to confirm that the device was added to the list instead of just double clicking on it (the picture below is an idea of where it could go)
Screen Shot 2022-07-25 at 12 13 40 PM
- each device has its own checklist
- each device signal within each checklist have the option to have their own comparisons

## Context
Pursuing the current route would be very tedious for the user and create more work than what atef is trying to eliminate the need of.

## Your Environment
source pcds_conda
my repo: /cds/home/l/liggett1/atef

### LINKED ISSUE #123: Refactor: restructure passive configurations to better reflect the most common use case
## Issues

The current configuration file format is fixed in its structure and flexible in ways that many users will not need (and worse, they find confusing and unintuitive).

Feedback has clued us in to the need for further restructuring(#121, #120, etc.), and after some discussion @ZLLentz and I came up with the following.

## The pitch

* Users will have more control over the hierarchy of configurations and checks being performed.
    * You will be able to sub-divide your configuration file into as many "Groups" as you would like. 
    * Each group may contain zero or more `{PV,Device,Tool}Configurations` or be further divided into Groups of its own.
    * For example, say if the top-level were divided into different groups associated with a given area, you might do one of the following:
        * Under each "area" Group, have a set of DeviceConfiguration instances for that area
        * Under each "area" Group, separate it further into groups such as `sub-area -> device-function -> device`
    * The user-configured hierarchy will be reflected in the configuration GUI, generated reports, and of course `atef check` output
* Many-to-many checks will remain supported.  
    * First-class support for 1:1 (or 1:many) device-signal-to-comparison(s) will be added.
    * First-class support for 1:1 (or 1:many) PV to comparison(s) will be added.

## "Groups"

Each group:
* Will be named with an optional description
* May store values for reuse in comparisons nested beneath
* May have options regarding how to interpret results of nested comparisons/groups
* May have one or more groups or comparisons underneath it

## GUI

(More details to follow)
* The overview tree will likely become the primary interaction method for adjusting checks

## The path forward

* #111 may or may not get in prior to the refactor (before would be preferable but not strictly required)
* The underlying `atef.config` needs reworking first and foremost
* GUI updates will follow
* A small migration script for old configuration files may be considered (no guarantee; and it will not be included in the repository)

klauer:
## Description

* Users will have more control over the hierarchy of configurations and checks being performed.
    * You will be able to sub-divide your configuration file into as many "Groups" as you would like. 
    * Each group may contain zero or more `{PV,Device,Tool}Configurations` or be further divided into Groups of its own.
    * For example, say if the top-level were divided into different groups associated with a given area, you might do one of the following:
        * Under each "area" Group, have a set of DeviceConfiguration instances for that area
        * Under each "area" Group, separate it further into groups such as `sub-area -> device-function -> device`
    * The user-configured hierarchy will be reflected in the configuration GUI, generated reports, and of course `atef check` output
* Each group may specify that the severity of the underlying configurations be minimized (`any`) or maximized (`all`) (terminology may need some work here)
* Many-to-many checks are supported alongside per-attribute checks
    * First-class support for 1:1 (or 1:many) device-signal-to-comparison(s) was added
    * First-class support for 1:1 (or 1:many) PV to comparison(s) was added
* If a 1-1 device:check is desired, it will be a separate `DeviceConfiguration` in a `ConfigurationGroup`
* Adds a conceptual `values` dictionary that can be configured at a `Group` level, which may be reused in dynamic device configurations (all a TODO)
* Adds a version number to the `ConfigurationFile` (am attempting to keep it somewhat flexible going forward)

## What's the catch?

* GUI has not been updated at all
* Dynamic value checks are still a separate, unmerged branch and need integration here
* Old configuration files will no longer be supported
* Need to re-add filtering for `atef check` for tags/device names/PV names/etc

## Motivation and Context
Closes #123 
Closes #121
Will close #120 when the GUI is updated

## How Has This Been Tested?
Test suite and locally

Primarily issue #123 and related discussions


klauer:
This is in a working state and does what we had planned. However:
1. The GUI will be out of sync with this and it will not pass tests. That work needs to be completed at some point.
2. The primary reviewers in this repo, @ZLLentz  and @tangkong, will be out of office for around a week

I think I'd rather wait until Zach and Robert have a chance to comment on the functionality here before the GUI rework gets done, so I'll probably jump to other projects in the meantime

atef/config.py
```diff
@@ -112,34 +127,46 @@ class ConfigurationFile:
     """
     A configuration file comprised of a number of devices/PV configurations.
     """
+    #: atef configuration file version information.
+    version: Literal[0] = field(default=0, metadata=apischema.metadata.required)
```
ZLLentz, line 131: How is this going to be used? Increment this number every time we update the schema? Use old schema info if we read a file with an old number?

Depending on what we want to do it might be appropriate to automate this via tagging

ZLLentz (commented):
Partial review- resuming tomorrow. This looks pretty good.

atef/config.py
klauer: If breaking changes are made in the schema, we bump the version.
Then we make a shim to convert older schemas to newer ones.
It may be possible to automate, but I think it might be more trouble than it's worth.

atef/config.py
```diff
@@ -53,8 +41,29 @@ class Configuration:
...
+
+@dataclass
+class ConfigurationGroup(Configuration):
+    """
+    Configuration group.
+    """
+    #: Configurations underneath this group.
+    configs: List[Configuration] = field(default_factory=list)
+    #: Values that can be reused in comparisons underneath this group.
+    values: Dict[str, Any] = field(default_factory=dict)
```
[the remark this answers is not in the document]
klauer: From the verbose PR text:
> Adds a conceptual values dictionary that can be configured at a Group level, which may be reused in dynamic device configurations (all a TODO)

I'm thinking that you could define values at the configuration group level and reuse them in tests underneath. It's more a proposal of something I thought would be useful than anything else. If you think it'd be useful, let's leave it in; if not, let's nuke it before this PR goes through.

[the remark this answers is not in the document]
ZLLentz: I have no idea if it'll be useful or not- maybe I just need to see a specific use case. I re-read the verbose PR text before I made the above comment.

Is this basically setting a global variable for a group configuration? e.g.
```
full_vacuum = 1e-10
...
vgc_02.pressure < full_vacuum
```
I could see it getting some use if it's something like that

atef/config.py
ZLLentz: `breaking changes` was the keyword I was missing that makes this useful/sustainable to update

atef/config.py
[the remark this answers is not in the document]
klauer: Correct. The idea would be that you would only have to change the value in one place instead of (e.g) 10.

The scoping rules would be such that the parent group values have precedence over the grandparent group values (and any ancestor groups). Values could come from any ancestor if undefined on the parent group.

That means you could have a set of values defined in the root of the configuration that get reused throughout

.travis.yml
```diff
@@ -27,4 +27,4 @@ jobs:
     - name: "Python - PIP"
 
 import:
-  - pcdshub/pcds-ci-helpers:travis/shared_configs/standard-python-conda-latest.yml
+  - klauer/pcds-ci-helpers:travis/shared_configs/standard-python-conda-latest.yml@ref_pytest-cov
```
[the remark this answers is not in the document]
klauer:
```suggestion
  - pcdshub/pcds-ci-helpers:travis/shared_configs/standard-python-conda-latest.yml
```

atef/config.py
```diff
@@ -157,14 +184,820 @@ def from_yaml(cls, filename: AnyPath) -> ConfigurationFile:
...
+            cache=cache,
+            parent=None,
+        )
+        prepared_file = PreparedFile(
+            file=file,
+            cache=cache,
+            client=client,
+            root=prepared_root,
+        )
+        prepared_root.parent = prepared_root
```
ZLLentz, line 254: This bit is unintuitive to me- isn't this inviting some sort of infinite loop accident? What's the advantage of setting its parent to itself rather than keeping it as `None` or some other sentinel value?

atef/config.py
```diff
@@ -157,14 +184,820 @@ def from_yaml(cls, filename: AnyPath) -> ConfigurationFile:
...
+                        )
+                    )
+                except Exception as ex:
+                    prepare_failures.append(ex)
+
+        return prepared
+
+
+@dataclass
+class PreparedToolConfiguration(PreparedConfiguration):
```
ZLLentz, line 893: It's a bit unfortunate that there has to be three extremely similar copies of these prepared configuration subclasses

ZLLentz (commented):
I finished my line-by-line, it looks good- just making sure there isn't anything small missed. I'm going to go back now with the files instead of the diff and do a bit of high-level review/thinking/design work about how some of this should look in the GUI.

atef/config.py
klauer: Unintuitive as it's a typo/bug - it should be `prepared_file`. The root configuration group's parent is the file.

atef/config.py
klauer: If you have suggestions as to how to restructure it, I'd be happy to revisit it

atef/config.py
klauer: Thanks for catching it - resolved in [40844c6](https://github.com/pcdshub/atef/pull/126/commits/40844c6984898603523778210afdcd17ed509760)

Test suite needs some additional group tests, I think.

atef/config.py
ZLLentz: I do not have any specific suggestions! I was just lamenting a bit

klauer: No problem, I thought it was unfortunate while writing it, too...

ZLLentz:
The test suite fails on test setup/widget imports- maybe it's prudent to at least update these such that the suite runs on CI? (Even if the ui itself will need an overhaul). Or maybe we should just comment out the failing imports in the test suite?

I think this looks good and I think there's a clear path for integrating dynamic checks + the gui on top of this. I can't immediately think of any major structural change to suggest, and this matches our offline discussion. The user gets a lot of flexibility here to organize their files and there's some hidden potential here for combining/splitting configuration files using the grouping mechanisms- any node that is a `ConfigurationGroup` could theoretically become its own file or be joined with other groups.

Some of the existing ui screens will want some extensions, e.g. comparisons will want to be able to show you which PVs/signals they'll be run on with the introduction of `by_attr` vs `shared`, among other things. The top-level object being itself a group config leaves a natural path for setting all of this up without a huge shakeup- just add some fields/modifications to the top-level screen and re-use it for all group layers.

The dynamic checks will probably need to initialize themselves at runtime during `PreparedComparison.get_data_async` which will probably want to mutate the `Comparison`, which should be ok as formulated.

ZLLentz:
Unrelated: should we put up a `v0.0.1` tag prior to this PR? The structural changes are pretty large and we're dropping file compatibility.

klauer:
> The test suite fails on test setup/widget imports- maybe it's prudent to at least update these such that the suite runs on CI? (Even if the ui itself will need an overhaul). Or maybe we should just comment out the failing imports in the test suite?

I'm giving that a go now, with some difficulty. I broke our entire CI pipeline 🥳 

> I think this looks good and I think there's a clear path for integrating dynamic checks + the gui on top of this. 

Great! Glad to hear it. We can chat about specifics if there's anything still confusing of course, too.

> Unrelated: should we put up a `v0.0.1` tag prior to this PR? The structural changes are pretty large and we're dropping file compatibility.

I think that's a great idea

ZLLentz:
https://github.com/pcdshub/atef/releases/tag/v0.0.1
My take at an alpha/pre-release tag, feel free to edit and we can discuss

klauer:
Looks like there were at least a couple remaining fixes required:
* happi client being unavailable in the test suite was breaking things (addressed in 7232cd997594df5879f932e0650dc3d4817ddd00)
* CLI verbosity settings were not showing the default with no CLI args (393a43ede9a4b6c2731f5c5c9af0f0b2cec2f041)
* Failed configurations were not showing in the tree (b369e4aa95dbd9b4bfc6a275027ed8f5171db6f6)
    * As in the device-based one which uses at2l0

So, breaking my local happi config intentionally results in:
```
$ HAPPI_CFG=/tmp atef check atef/tests/configs/device_based.yml
WARNING:atef.util:Unable to load happi Client from configuration (RuntimeError): Unable to instantiate the client. Please verify that your HAPPI_CFG points to the correct file and has the required configuration settings. In '/tmp', found settings: {}.
atef/tests/configs/device_based.yml
└── ❌Error: Root
    └── ❌: Failed to instantiate configuration: Failed to load happi device: at2l0. Configuration is: Basics ('')
```

klauer:
Looking back on this:
* Tests passing on CI now
* We have our pre-"massive change tag" as mentioned above

Given this is backward incompatible with previous versions and breaks the GUI, should we:

1. Merge and open a new PR with the GUI changes
2. Push in GUI updates here

ZLLentz:
If the goal is to keep master "working", then we could merge this into a new non-master side branch until the GUI catches up, maybe?

klauer:
> If the goal is to keep master "working",

I'm not sure it is - do we care?

ZLLentz:
I certainly don't care. I'm happy to start working on the ui.

ZLLentz:
Did I never put in my green checkmark somehow what

ZLLentz (approved):

klauer:
Here goes nothing...

Issue #34, opened 2022-04-19
Clean up: remove unused .ui files from atef config


ZLLentz:
## Expected Behavior

Some of these files were scrapped and I've lost track of which ones. We can grep for their titles and delete the unused ones.

PR #127, opened 2022-10-04
ENH: catch up the GUI to the rest of the repo

### LINKED ISSUE #30: Feature: Implement all Comparison widgets for atef config GUI

### LINKED ISSUE #80: BUG: tiny scrollbar in various description fields
## Expected Behavior

Description textboxes shouldn't have a scrollbar at one line- it should be no scrollbar at all (expanding for each new line) or it should be scrollbar after N lines (N in 3-5?)

## Current Behavior

I noticed this specifically when opening a display from a file:

image

### LINKED ISSUE #92: Comparisons Summary on Id and Comp Widget
## Expected Behavior

Comparison listing: a summary in a tooltip or a widget to the side might be nice

_Originally posted by @klauer in https://github.com/pcdshub/atef/issues/71#issuecomment-1151378962_

### LINKED ISSUE #98: Can't delete configurations unless the name is deleted first
## Expected Behavior
Simply pressing the "Delete Configuration" button should remove the entry without having an empty name first

## Current Behavior
User has to remove the configuration name first, then delete

1. Find a test config to delete
2. Press delete button and nothing happens
3. Delete entry name
4. Delete entry now works

## Context

It's inconvenient but not a priority

## Your Environment

source pcds_conda
My repo at /cds/home/c/cpino/atef
Andrew's repo at /cds/home/l/liggett1/atef

### LINKED ISSUE #104: Effectiveness of "Lock/Unlock Editing" Button
## Expected Behavior
When activated, the user should not be able to change anything about the configuration.

## Current Behavior
There is a loophole to this by selecting what to edit from the side menu, bypassing the lock/unlock editing button

## Possible Solution
Have the button affect the whole configuration, no matter what screen is open.

Perhaps adding a password protection to unlock the editing.

1. When creating a PV config, click on the configuration name in the side menu.
2. This will bring you to the checklist page, which will also allow you to edit the name and description of the PV.

## Context
This problem essentially bypasses the whole button in general and prevents it from being used as originally intended.

## Your Environment
source pcds_conda
My repo: /cds/home/l/liggett1/atef

### LINKED ISSUE #124: Enhancement: add tool configurations to the atef config GUI
## Expected Behavior
Tools from #114 should be added to the `atef config` GUI to allow for more user-friendly tool configuration specification.

`ToolConfiguration` will need to be supported as well as the individual `Tool` subclasses, of which there is only currently one: `Ping`.

## Current Behavior
Tools are not in the GUI

## Possible Solution
Add tools to the GUI

1. Start `atef config`
2. See no tools
3. Frown

## Context
Added in #114

ZLLentz:
## Description

- Follow-up to #126 to allow us to make configs using the GUI again
- Refactor the code organization heavily
- Various usability improvements
- closes #124 (adds tools)
- closes #104 (removes lock button)
- closes #98 (x always works, asks for confirmation)
- closes #92 (adds helpful text in top-right of comparisons pages)
- closes #80 (bugfix)
- closes #30 (no missing data types or fields now)
- may close others but I'm not sure
- Update lfe demo config
- Add full(?) coverage demo config

Warning: the code here is much messier than my usual output. I started rushing through this once I realized how long it was taking (GUIs are hard).

TODO:

- [ ] address #34
- [ ] Code deduplication
- [ ] Missing docstrings
- [ ] Missing annotations
- [ ] Unit test(s)

## Motivation and Context

The GUI had fallen behind the feature set of the passive checks and needed to catch up.
The previous organization made it hard to figure out what each widget was supposed to do.
I didn't do a great job of fully separating data manipulation from tree manipulation, but it's more manageable now than before.

## How Has This Been Tested?

Interactively only, I intend to add (a) unit test(s)

Not yet, I guess a tutorial could be added

image


atef/ui/any_value_widget.ui
```diff
@@ -0,0 +1,98 @@
...
+    <number>0</number>
+   </property>
+   <property name="rightMargin">
+    <number>0</number>
+   </property>
+   <property name="bottomMargin">
+    <number>0</number>
+   </property>
+   <item>
+    <widget class="QTableWidget" name="values_table">
```
klauer, line 30: Would recommend this setting for these table widgets for a low-effort way to fill the space a bit better:
image

atef/widgets/config/page.py
```diff
@@ -0,0 +1,1664 @@
...
+
+        Shoutouts to stackoverflow
+        """
+        if event.source() is self.config_table:
+            selected_indices = self.config_table.selectedIndexes()
+            if not selected_indices:
+                return
+            selected_row = selected_indices[0].row()
+            dest_row = self.config_table.indexAt(event.pos()).row()
+            if dest_row == -1:
```
klauer, line 524: I was testing the neat drag and drop functionality and found that for a simple one like `demo-configs/lfe.json`, dragging one of the configs to the end of the table caused it to disappear. 

Seems like the row count may be off-by-one, making it show up past the visible number of rows. For this single case, this worked:
```python
            if dest_row == -1:
                dest_row = self.config_table.rowCount() - 1
```
But hmm, it does make me wonder if `dest_row > source_row`, `dest_row` should be decremented by one (to account for the removed row) in general?

ZLLentz: > dragging one of the configs to the end of the table caused it to disappear

oh no I thought I tested this thoroughly

ZLLentz: I originally had some decrement logic in for dest_row > source_row but it seemed to break the table. When I removed it I thought I had everything working but I guess it's time to revisit.

klauer: Hmm, strange...

ZLLentz: The drag and drop feels pretty robust with the tweak you suggested above. I don't fully understand why the decrement isn't needed for the dest > source case but it's definitely clear why placing the widget into a row outside the bounds of the table would break things- the last row is row count - 1, not row count. I guess when I was originally testing this I didn't have the table sized to be big or didn't drag into the blank space below the bottom row of the table, since all the drags within the bounds of the rows did work without subtracting 1 here.

klauer: Sounds good enough to me! Thanks for double-checking the other scenario

atef/ui/any_value_widget.ui
ZLLentz: This ends up being a pretty major simplification for a lot of my extra event handlers/row spacing and gives a better result at the end

ZLLentz:
I don't think I actually have time this month to fully clean up this PR. I think I'm going to focus on docstrings/annotations and write one test and then move on to the next project.

ZLLentz:
While writing out the docs/annotations I noticed that there are a few low-hanging-fruit things to address that I'll do before asking for review- but by and large I'm not going to completely generalize all the code during this PR and some things will still be messy.

ZLLentz:
This is at an "I've left this open for too long" sort of phase more than anything else

klauer (approved):
This works well in my testing. 

While we'll undoubtedly find minor issues here and there, we can fix them in follow-up PRs.

tangkong (approved):
This also works well for me in testing.  I have a variety of follow-up thoughts, but I don't think any of them should block the merging of this PR.  I think it does what you set out to do.  

Some thoughts I'll toss into new issues:
- Closing behavior
  - Should it be possible to close a file without closing the whole window?
  - Should we confirm unsaved changes before closing the main window?
  - We can currently close the main window while the happi/ophyd selector window stays open
- I imagine enforcing types would be good for fields like reduce period or range comparisons
- Could we put tooltips on the common settings (tolerances/reduce periods) in case users want to know exactly what the math is?

It's possible you've both got these in your sights already, or just punted the issues.  These seem like low-ish hanging fruit for qt newbies like me to tackle

ZLLentz:
Some of those I noticed but forgot about/punted, others I didn't notice/didn't consider. They are all good ideas, some bordering on essentials.

Issue #69, opened 2022-05-03
Feature: Archive Helper in Happi Component Selection


ZLLentz:
## Expected Behavior

There should be a way to check historical data for a particular signal when you're selecting it to help you design your test.

## Current Behavior

No such option

## Context

My interpretation of something Alex mentioned during our quick 1-on-1 catchup meeting when I showed him some atef progress

tangkong:
Closed by #132 , a while ago

PR #132, opened 2022-11-01
ENH: add archive viewer widget

### LINKED ISSUE #130: BUG/GUI: right-click context menu in Happi widget shows duplicate attributes
## Expected Behavior
Context menu should only show a single attribute
## Current Behavior
image
The "Add Single Attribute" section shows duplicate entries

1. open the Happi/Ophyd signal selection menu
2. select a device 
3. right click on one of the individual signals

## Context
Was poking around and stumbled across this

## Your Environment
master + zach's big gui pr

tangkong:
## Description
Adds an archive appliance viewer widget to help during check creation.  Leverages PydmArchiverTimePlot

Includes neat features (read: time sinks) like:
- a neat table view that holds pv curve information (symbol, color, etc.  More to come)
- basic `archapp.EpicsArchive` validation on PV input
- integration with existing context Ophyd/Happi device context menus 

## Motivation and Context
#69 
closes #130 along the way

## How Has This Been Tested?
interactively 

Lots of comments as I learn more about qt
this PR. 

(Opening from the happi/ophyd component selector)
image

(opening from main window)
image

(opening from comparison page, opens all signals related to the comparison)
image

image

Double clicking on items in the table brings up edit windows
color picker: 
image

symbol combobox
image

### DIFF
--- a/atef/widgets/archive_viewer.py
+++ b/atef/widgets/archive_viewer.py
@@ -4,11 +4,13 @@ Widget classes designed for PV archiver interaction
 
 from __future__ import annotations
 
+import datetime
 import logging
 import os
 import urllib
 from typing import Any, ClassVar, Dict, List, Optional
 
+from archapp.interactive import EpicsArchive
 from pydm.widgets.archiver_time_plot import PyDMArchiverTimePlot
 from qtpy import QtCore, QtGui, QtWidgets
 from qtpy.QtCore import QRegularExpression
@@ -105,8 +107,14 @@ class ArchiverViewerWidget(DesignerDisplay, QWidget):
             archiver_url = get_pingable_url(ARCHIVER_URLS)
             if archiver_url is None:
                 raise ArchiverError('Cannot reach any archiver urls')
-            print(f'setting archiver url to: {archiver_url}')
+            # need to set environment variable for archiver data plugin
+            logger.debug(f'setting archiver url to: {archiver_url}')
             os.environ['PYDM_ARCHIVER_URL'] = archiver_url
+        else:
+            archiver_url = os.environ['PYDM_ARCHIVER_URL']
+
+        url_core = archiver_url.removeprefix('http://').split('.', 1)[0]
+        self.archapp = EpicsArchive(url_core)
 
         self._pv_list = pvs
         for pv in pvs:
@@ -184,6 +192,16 @@ class ArchiverViewerWidget(DesignerDisplay, QWidget):
             # grab and clear text
             pv = self.input_field.text()
 
+            # check if data exists
+            data = self.get_pv_data_snippet(pv)
+            if len(data['data']) < 3:
+                QtWidgets.QMessageBox.information(
+                    self,
+                    'Invalid PV',
+                    '< 3 datapoints from last two days found in '
+                    'archiver app, skipping add'
+                )
+
             # add item
             self.add_signal(pv)
 
@@ -195,18 +213,22 @@ class ArchiverViewerWidget(DesignerDisplay, QWidget):
 
         self.time_plot.clearCurves()
         for pv in pv_data:
+
+            data = self.get_pv_data_snippet(pv[0])
+
             self.time_plot.addYChannel(
                 y_channel=f'ca://{pv[0]}',
-                name=f'{pv[0]}',
+                name=f'{pv[0]} ({data["meta"].get("EGU","")})',
                 symbol=pv[1]['symbol'],
                 color=pv[1]['color'],
                 lineStyle=pv[1]['lineStyle'],
                 lineWidth=pv[1]['lineWidth'],
-                useArchiveData=True
+                useArchiveData=True,
+                yAxisName='yAxis'
             )
 
         try:
-            self.time_plot.setLabel('left', text='')
+            self.time_plot.setLabel('yAxis', text='')
         except Exception:
             # pyqtgraph raises a vanilla exception
             # if a better way to find the left axis name exists, use it
@@ -219,6 +241,15 @@ class ArchiverViewerWidget(DesignerDisplay, QWidget):
         if success:
             self._update_curves()
             self.input_field.clear()
+
+    def get_pv_data_snippet(self, pv: str) -> Dict[str, Any]:
+        # use raw get for json metadata
+        # also sidesteps an issue where some PV's aren't found using
+        # the normal EpicsArchive.get()
+        today = datetime.datetime.today()
+        prev = today - datetime.timedelta(days=2)
+        data = self.archapp._data.get_raw(pv, prev, today)
+        return data
 
 
 class PVModel(QtCore.QAbstractTableModel):
--- a/atef/widgets/archive_viewer.py
+++ b/atef/widgets/archive_viewer.py
@@ -48,11 +48,12 @@ def get_archive_viewer() -> ArchiverViewerWidget:
 def get_pingable_url(urls: List[str]) -> str:
     """
     Get valid archiver URLS from the urls
+    Looks only for an response code below 400, as a proxy ping test.
 
     Returns
     -------
     str
-        a valid archiver url
+        a responsive url
     """
     for url in urls:
         try:
@@ -72,15 +73,12 @@ class ArchiverError(Exception):
 
 # TO-DO: crosshair hover over plot?
 #   -> time_plot.enableCrosshair
-# TO-DO: scaling / dealing with different scales
-# TO-DO: set the units in the label? maybe ignore it, nothing else has it
-# TO-DO: set up validator on QLineEdit, PV exists, has data?
+# TO-DO: scaling / dealing with different scales between curves
 # TO-DO: Turn off autoscale.  Set to 1 day automatically
+# TO-DO: make buttons work after manual zoom
 # TO-DO: Make redraw retain scale.
-# TO-DO: get rid of bars
-#   - to reproduce
-#       - plot, zoom out to fill data, then redraw.  bars will remain
-#   - not plotstyle?...
+# TO-DO: Cycle colors on curve add
+# TO-DO: get rid of bars --> Known bug with PydmArchiveTimePlot
 class ArchiverViewerWidget(DesignerDisplay, QWidget):
     """
     Archiver time plot viewer
@@ -146,7 +144,6 @@ class ArchiverViewerWidget(DesignerDisplay, QWidget):
 
         # delete button in last column
         self.deleteDelegate = DeleteDelegate()
-        # -1 doesn't work sadly
         del_col = len(self.model.headers) - 1
         self.curve_list.setItemDelegateForColumn(del_col, self.deleteDelegate)
         self.deleteDelegate.delete_request.connect(self.model.removeRow)
@@ -158,7 +155,6 @@ class ArchiverViewerWidget(DesignerDisplay, QWidget):
         self._setup_pv_selector()
 
         self.redraw_button.clicked.connect(self._update_curves)
-        # pass
 
     def _setup_range_buttons(self):
         def _set_time_span_fn(s: float):
@@ -172,7 +168,10 @@ class ArchiverViewerWidget(DesignerDisplay, QWidget):
                 The time span in seconds
             """
             def fn():
-                self.time_plot.setTimeSpan(s)
+                now = datetime.datetime.today()
+                prev = now - datetime.timedelta(seconds=s)
+                self.time_plot.setMaxXRange(now.timestamp())
+                self.time_plot.setMinXRange(prev.timestamp())
                 self.time_plot.updateXAxis()
             return fn
 
@@ -181,7 +180,6 @@ class ArchiverViewerWidget(DesignerDisplay, QWidget):
         self.button_month.clicked.connect(_set_time_span_fn(24*60*60*30))
 
     def _setup_pv_selector(self):
-        # inputs get made into list items
         # validate PV form, returnPressed unless valid
         regexp = QRegularExpression(r'^\w+(:\w+)+(\.\w+)*$')
         validator = QRegularExpressionValidator(regexp)
@@ -198,9 +196,10 @@ class ArchiverViewerWidget(DesignerDisplay, QWidget):
                 QtWidgets.QMessageBox.information(
                     self,
                     'Invalid PV',
-                    '< 3 datapoints from last two days found in '
+                    'Fewer than 3 datapoints from last two days found in '
                     'archiver app, skipping add'
                 )
+                return
 
             # add item
             self.add_signal(pv)
@@ -211,6 +210,7 @@ class ArchiverViewerWidget(DesignerDisplay, QWidget):
         # grab all the list items
         pv_data = self.curve_list.model().pvs
 
+        # Re-add curves
         self.time_plot.clearCurves()
         for pv in pv_data:
 
--- a/atef/widgets/archive_viewer.py
+++ b/atef/widgets/archive_viewer.py
@@ -1,0 +1,149 @@
+"""
+Widget classes designed for PV archiver interaction
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import ClassVar, List, Optional
+
+from pydm.widgets.archiver_time_plot import PyDMArchiverTimePlot
+from qtpy import QtWidgets
+from qtpy.QtCore import QRegularExpression
+from qtpy.QtGui import QRegularExpressionValidator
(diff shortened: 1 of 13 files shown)


tangkong:
Test failures stem from an updated pcdsdevices.interface.LightpathMixin, unrelated to this PR #133

klauer:
Neeeeat

atef/widgets/config/utils.py
```diff
@@ -373,6 +374,16 @@ def add_with_range():
...
             self.suggest_comparison.emit(comparison)
 
+        def open_arch_viewer():
+            arch_widget = get_archive_viewer()
+            for datum in data:
+                arch_widget.add_signal(datum.pvname)
+            arch_widget.show()
+
+        menu.addSection("Open Archive Data viewer")
+        archive_viewer_all = menu.addAction("View all selected")
```
klauer, line 384: I think it's worthwhile to have a prominent additional button between 'select' and 'update data' in the happi component selection dialog.

klauer (commented):
This is super cool. atef is turning into such a nice searching/viewing tool for everything in the facility.

atef/widgets/archive_viewer.py
```diff
@@ -0,0 +1,411 @@
...
+from qtpy import QtCore, QtGui, QtWidgets
+from qtpy.QtCore import QRegularExpression
+from qtpy.QtGui import QRegularExpressionValidator
+from qtpy.QtWidgets import QWidget
+
+from atef.widgets.core import DesignerDisplay
+
+logger = logging.getLogger(__name__)
+archive_viewer_singleton = None
+ARCHIVER_URLS = ['http://pscaa01.slac.stanford.edu',
```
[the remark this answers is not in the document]
tangkong: Notes for posterity:  
- I contacted Omar and he brought pscaa02 back online... somehow.  I'm not sure how long-lived that will be, so I'm inclined to leave this bit of logic in there.
- I'm actually unclear as to how `archap.EpicsArchive` works with the little stub of a url.  For now the ArchiveTimePlot widget expects fully formed url, and will throw QSslSocket errors if we try with the shortened path.  `EpicsArchive` seems to work with a more fully specified hostname (eg `pscaa01.slac.stanford.edu`) so I've unified things behind that.  Unfortunately pyDM needs the port specified while `EpicsArchive` adds it itself, so there's still that difference.
- If you specify the `PYDM_ARCHIVER_URL` env var, it should use that.  Though I still make assumptions about the port number..
- Names are hard

atef/widgets/config/utils.py
tangkong: What kind of button are you thinking of?  Right now the archive viewer button will open the viewer widget if it doesn't exist or add pv's the widget if it does.

atef/widgets/archive_viewer.py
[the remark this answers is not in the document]
klauer: > Unfortunately pyDM needs the port specified while EpicsArchive adds it itself, so there's still that difference.

This definitely makes using the same configuration setting awkward at best and broken at worst (though only for non-standard port 17668 configs). 

Archapp appears to always append the port, even if it's already there: https://github.com/pcdshub/archapp/blob/3544d9eb94f56ae2635a0fa73e12ed67bd38a242/archapp/url.py#L49

Not sure what the actual best approach would be, but the best I can think of at the moment is to use the PyDM environment variable and remove the port for `archapp`, raising if it's not 17668 or something? This is likely extending beyond the scope of this PR though it is related enough to be discussed here, I think.

atef/widgets/config/utils.py
klauer: Hmm, maybe I'm missing the button. Could you include a screenshot? This is the context menu where I've been accessing the archive viewer plots from:
image
I was thinking a button in between "Select" and "Update data" would be nice

tangkong: Interesting.  For me it renders like this, with context menu section headers:
image

But I realize now you're thinking of a button outside of the context menu that will open the archive viewer.  That's something I can add 👍

klauer: Yeah, those section headers are OS dependent, frustratingly enough.

atef/widgets/archive_viewer.py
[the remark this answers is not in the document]
tangkong: > but the best I can think of at the moment is to use the PyDM environment variable and remove the port for archapp, raising if it's not 17668 or something?

This is close to what I've got so far, I'll finish it up in that vein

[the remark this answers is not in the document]
ZLLentz: `archapp` is a very old beta product and can be treated as such. We can change whatever we want if any decisions made in that codebase are disruptive.

atef/widgets/archive_viewer.py
```diff
@@ -0,0 +1,622 @@
...
+        # set up table view for PV info
+        self.model = PVModel(parent=self)
+        self.curve_list.setModel(self.model)
+        horiz_header = self.curve_list.horizontalHeader()
+        horiz_header.setSectionResizeMode(horiz_header.ResizeToContents)
+
+        # set up delegates
+        # Color picker delegate
+        self.colorDelegate = ColorDelegate()
+        self.curve_list.setItemDelegateForColumn(2, self.colorDelegate)
```
ZLLentz, line 178: I wish I had picked up on these delegates when I was taking over `pmps-ui`, it's a much cleaner way to handle these tables

tangkong: It took me longer than I wanted to sort them out, but they do turn out quite nicely once they're working 👍

atef/widgets/archive_viewer.py
```diff
@@ -0,0 +1,622 @@
...
+                    logger.warning(f'PYDM_ARCHIVER_URL ({archiver_url}) does '
+                                   'not end with port 17668, replacing port')
+                    archiver_url = archiver_url.replace(port_re[0], ':17668')
+                else:
+                    archiver_url += ':17668'
+                os.environ['PYDM_ARCHIVER_URL'] = archiver_url
+
+        # EpicsArchive wants a stripped down url
+        url_core = archiver_url.removeprefix('http://').removesuffix(':17668')
+        self.archapp = EpicsArchive(url_core)
```
ZLLentz, line 151: Not sure where to put this- but it seems like it's possible to "outrun" the archiver gets by impatiently clicking buttons and cause the rendering to fail.

atef/widgets/config/utils.py
```diff
@@ -373,6 +374,18 @@ def add_with_range():
...
             self.suggest_comparison.emit(comparison)
 
+        def open_arch_viewer():
+            arch_widget = get_archive_viewer()
+            for datum in data:
+                dev_attr = '.'.join((datum.signal.parent.name, datum.attr))
+                arch_widget.add_signal(datum.pvname, dev_attr=dev_attr)
+            arch_widget.show()
+
+        menu.addSection("Open Archive Data viewer")
```
ZLLentz, line 384: I notice that these are hooked in to the object/component chooser specifically.

1. Any thoughts on how we could include it in the PV configurations too?
2. My intuition is that the most useful place to hook into the archiver is actually in the comparison editor- knowing that the gauge dips below x value when things are bad is most useful when you're trying to set up your greater-than comparison, for example.

atef/widgets/archive_viewer.py
ZLLentz:
```
  File "/cds/home/z/zlentz/miniconda3/envs/pcds-5.5.1/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/cds/home/z/zlentz/miniconda3/envs/pcds-5.5.1/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/cds/home/z/zlentz/github/atef/atef/__main__.py", line 3, in <module>
    main()
  File "/cds/home/z/zlentz/github/atef/atef/bin/main.py", line 99, in main
    func(**kwargs)
  File "/cds/home/z/zlentz/github/atef/atef/bin/config.py", line 46, in main
    app.exec()
  File "/cds/group/pcds/pyps/apps/dev/pythonpath/pydm/widgets/archiver_time_plot.py", line 112, in receiveArchiveData
    self.getViewBox().addItem(self.error_bar_item)
AttributeError: 'NoneType' object has no attribute 'addItem'
```

atef/widgets/config/utils.py
klauer: Agreed on both points. Expanding on that, perhaps:

It would be nice to have a 'view in archiver' button for the PV lists and/or context menu 'view in archiver' for individual PVs. The device configuration section could have something near the 'signals' section?
image image

It'd be nice to have the option in the comparison editor itself as well, a button up near the top? A drop-down QToolButton menu for "additional actions"?
image

atef/widgets/archive_viewer.py
ZLLentz: I was always able to reset and get something after clicking more patiently. I'm not sure if anything can be done to make it obvious that the gui is waiting for an archiver response.

tangkong: I'm starting to suspect this to be more of a pydm side bug than something I can fix here...  This can more easily be replicated if you load a few PV's and increase the range to 1 month or something large.  Before all three can grab archive data, the plot will complain.  The data loads just fine afterwards though

ZLLentz: It's absolutely possible that it's a pydm bug

ZLLentz:
First, I think the features here are great and an important addition to atef. Here's all the weird stuff I saw from clicking around:
- Clicking the signal archiver button from the devices config always throws an exception. This is true for the right click context menu and for the button at the bottom. 
image
- The comparison page archiver button works great in the cases I tried, but has a false error. Starting from a cleared archiver widget I keep getting pop-ups that indicate some sort of duplicate PV warning but then everything works as it should regardless: 
image
- the tools comparison page probably shouldn't have an archiver button, I think it shouldn't be too hard to locate and hide it
- the pydm archiver widget needs a hero some day to smooth out its rough edges
- the comparison page layouts get a little weird with the extra space stuck into them and I can't put my finger on what exactly is happening to them. Not a big deal of course. 
image

atef/widgets/archive_viewer.py
```diff
@@ -0,0 +1,619 @@
...
+        Parameters
+        ----------
+        pv : str
+            the PV to be added (eg. MR2L0:RTD:1:TEMP)
+        dev_attr : Optional[str], optional
+            the ophyd attribute name corresponding to the ``pv``,
+            by default None
+        """
+        # check if data exists
+        data = self.get_pv_data_snippet(pv)
```
klauer, line 295: I clicked on the awesome new "view archive data" button in the PV configuration comparison of `lfe.json`, and it's doing what I've asked. However, the archive requests take a long time and are run in the GUI thread, meaning it blocks everything else happening in atef until these things are run.  Perhaps it should be sent to a thread, instead? 

In the case of a slow archiver response over ssh, this is actually taking well over a minute (maybe two at this point? I started typing the comment when I noticed it, and it still hasn't finished)... 

Edit: hmm, it's still going. I have a slight modification of archapp that [indicates this](https://gist.github.com/klauer/533197dbf3698353beb7bfed682a17ad) for a single click. That seems like way too many requests. What's going on here?

atef/widgets/archive_viewer.py
```diff
@@ -0,0 +1,619 @@
...
+    """
+    filename: ClassVar[str] = 'archive_viewer_widget.ui'
+
+    time_plot: PyDMArchiverTimePlot
+    button_month: QtWidgets.QPushButton
+    button_week: QtWidgets.QPushButton
+    button_day: QtWidgets.QPushButton
+    input_field: QtWidgets.QLineEdit
+    curve_list: QtWidgets.QTableView
+    redraw_button: QtWidgets.QPushButton
```
klauer, line 117: A "clear curves" button might also be nice, for when you want to start from scratch

atef/widgets/archive_viewer.py
tangkong: This is weird... I only get one call to the archapp `get_raw` method whenever I draw a curve, as I expect.  That was for a single click of which button?  The "view archive data" button on the PV comparison page?

Your output definitely shows too many requests (multiple calls to a single PV), but I can't reproduce it from psbuild.  I wonder if archapp is retrying failed requests... it doesn't seem to be from my reading of the code...

All that being said it's probably worth tossing in a thread.  👍

klauer: archapp v1.1.0
`python -m atef config atef/tests/configs/lfe.json`
On "Good vacuum", click the archive viewer button:
image
This may be retries-related. Some of this may still be a me-problem, as `curl` -ing those URLs locally works without issue, but archapp/PyDM seem to take issue with them

tangkong:
> Clicking the signal archiver button from the devices config always throws an exception. This is true for the right click context menu and for the button at the bottom.

aaah I forgot about `GroupDevice` and their lack of parents

> The comparison page archiver button works great in the cases I tried, but has a false error. Starting from a cleared archiver widget I keep getting pop-ups that indicate some sort of duplicate PV warning but then everything works as it should regardless:

Is it possible that you double-clicked the comp-page archiver button?  I can't reproduce this slowly, but there seems to be no lockout on the button, so if you double-click it, the pv's will be added to the viewer twice with the second set throwing the duplicate PV warning

> the comparison page layouts get a little weird with the extra space stuck into them and I can't put my finger on what exactly is happening to them. 

I played around with the word wrap and spacers, but didn't find anything I really loved.  I ended up placing a minimum width on the name box so it wouldn't squeeze itself out of existence.  This made the text on the description clip, but I duplicated the text into its tooltip so it's still accessible.  Totally open to alternate opinions here though

ZLLentz:
- I'm not double-clicking (I think?) but it might be related to the group device again?
- The tooltip is great and it's not worth spending lots of time on the spacing of a non-button

atef/widgets/archive_viewer.py
tangkong: Ken astutely noticed that this was querying each PV N^2 times, where N was the number of PV's being added at once.  This was fixed by not updating the plot with every PV added, rather updating it all at the end

The calls to archapp were also cached.  [66a1923](https://github.com/pcdshub/atef/pull/132/commits/66a1923dbb540372fed0224df42369541443813b)

I'm looking into shoving the calls into worker threads now as well

tangkong:
The last thing I had on the docket here was to try and toss the archapp calls into a thread, in order to free up the gui while PV's were being added to the archive viewer.  I tried a few things, but ended up hitting upon the fact that python's [signal](https://github.com/pcdshub/archapp/blob/3544d9eb94f56ae2635a0fa73e12ed67bd38a242/archapp/url.py#L70) only works in the main thread of the main interpreter.  😞 

<details>
  <summary> traceback </summary>

```
  File "/cds/home/r/roberttk/devrepos/atef/atef/qt_helpers.py", line 332, in run
    self.return_value = self.func(*self.args, **self.kwargs)
  File "/cds/home/r/roberttk/devrepos/atef/atef/widgets/config/data.py", line 267, in add_sig
    widget.add_signal(pv, dev_attr=dev_attr, update_curves=False)
  File "/cds/home/r/roberttk/devrepos/atef/atef/widgets/archive_viewer.py", line 314, in add_signal
    data = self.get_pv_data_snippet(pv)
  File "/cds/home/r/roberttk/devrepos/atef/atef/widgets/archive_viewer.py", line 348, in get_pv_data_snippet
    data = self.archapp._data.get_raw(pv, prev, today)
  File "/cds/home/r/roberttk/devrepos/archapp/archapp/data.py", line 166, in get_raw
    data = get_json(url)
  File "/cds/home/r/roberttk/devrepos/archapp/archapp/url.py", line 70, in get_json
    signal.signal(signal.SIGALRM, _raise_timeout)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.5.1/lib/python3.9/signal.py", line 56, in signal
    handler = _signal.signal(_enum_to_int(signalnum), _enum_to_int(handler))
ValueError: signal only works in main thread of the main interpreter
```

</details>

perhaps a deeper rework of `archapp` might be relevant, but at least for now caching should make this less painful on repeat archive-viewer openings?

ZLLentz (approved):

tangkong:
Thanks for the reviews everyone.  I'll toss up an issue reminding us to revisit the threading idea if archapp allows it in the future.

PR #134, opened 2022-11-04
ENH: command-line converter tool for the old file format

### LINKED ISSUE #131: Create tool to convert old atef passive checkout configurations to "v1" ones
## Issue

* There are dozens of configuration files created for our proof-of-concept passive checkout format
* We need a tool to convert from those old configuration files to the new, supported format
    * This new format is versioned at the file-level [here](https://github.com/ZLLentz/atef/blob/1a075648931664e8149372372df54ee2f293fce8/atef/config.py#L131)

## Opinion

* I don't think this script should be a prominent part of the atef package, nor a part of the configuration GUI.
* I think it should be standalone and sit in perhaps `/scripts` until we convert and migrate all files
   * After which point, we remove the script from the repository
* I think that working at the JSON level instead of the dataclass level will be appropriate for the conversion tool
* We should check the tool against all [existing configuration files](https://github.com/pcdshub/las-atef-config) and ensure that their results load properly with atef master

klauer:
## Description
This script will convert a prototype atef configuration file to the latest supported (and numbered) version.

## Motivation and Context
Closes #131 

## How Has This Been Tested?
* Ran it on `las-atef-config` without issue.
* Loaded resulting configuration in the GUI successfully

(This may need some more testing)

Issue and PR

image

### DIFF
--- a/scripts/converter_v0.py
+++ b/scripts/converter_v0.py
@@ -10,7 +10,7 @@ import json
 import logging
 import pathlib
 from dataclasses import dataclass, field
-from typing import List, Optional, Union
+from typing import Dict, List, Optional, Tuple, Union, cast
 
 import apischema
 import yaml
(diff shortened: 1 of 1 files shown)


tangkong:
tests fail due to the thing I pointed out in the archiver viewer widget PR.  I'll be working on that a bit longer so I should just pull that out and PR it sooner than later.

klauer:
Given that this is a throwaway script useful for only a couple people, I think I don't want to add tests (especially since it's outside the `atef` package itself). Agree/disagree?

If we agree, then this is ready for review (@ZLLentz @tangkong )

tangkong (approved):
This looks good to me.  My first time seeing the old atef configuration.  

The one thing that confused me briefly was overloading of the `load` function, but that's more of a nitpick.  (and maybe an issue with me reading things out of order)

klauer:
> The one thing that confused me briefly was overloading of the load function, but that's more of a nitpick. (and maybe an issue with me reading things out of order)

Hmm, which part do you mean?

tangkong:
erm, nevermind I for some reason thought `load` was a built-in 🤦 ignore me

ZLLentz (approved):
Looks clean to me- if it works, it works, and I agree there's no need to support it indefinitely with tests etc.,

This is basically a copy/paste of the prototype dataclasses + small tweaks + a conversion mapping layer to the modern ones, right?

ZLLentz:
The test environment builds in a failing way because `lightpath` isn't on `conda-forge`, and therefore isn't in the run requirements for `pcdsdevices`

klauer:
> This is basically a copy/paste of the prototype dataclasses + small tweaks + a conversion mapping layer to the modern ones, right?

Correct. There could be some argument as to better ways to map single configs with multiple identifier+comparison lists onto the new structure, I think. I left it almost as simple as possible. If our users give feedback as to how they'd like it, we can tweak it at that point.

ZLLentz:
Yeah, I think it's a bit of a big ask to figure out when/where to map into a group config for example. This should work cleanly as-is.

PR #140, opened 2023-01-25
ENH: run mode and basic report output

tangkong:
## Description
- Adds a run mode to the gui, accessible via a toggle in the config gui tab
  - image --> image
- Adds Run-Mode to the GUI, which allows top-level configurations and  individual comparison steps to be run independently.
  - image
- Adds basic report generation, including a results summary, linked table of contents, and best-effort page generation.
  - image
  - image

### BONUS!
Travis refused to build pcdsdevices properly so this PR now comes with the GHA migration.

## Motivation and Context
What do we want? To run our checks in the gui and print reports!
When do we want it? Now?

The broad plan I have is to split efforts into the following PR's
- Add run functionality to the GUI, along with basic report printing framework (this PR)
  - Limit this PR's scope to passive checkouts
- Build out active checkout framework under the previous Edit-Run framework
  - Build Edit and Run widgets for various existing steps
  - ... and more

<details>
<summary>Some notes / thoughts:</summary>

- Getting run-functionality up for passive checkouts was prioritized partly due to the desire to keep the UI/UX unified.  Since we would be building an atef view that runs the active checks interactively, a similar system for the passive checks would be expected.  Also reusing the existing tree systems was enticing.
- Reportlab can be unwieldly, but seemed flexible and powerful.  Also allowed me to write minimal html
- `AnyComparison` is an edge case that I decided to entirely ignore for this PR.  The problems it causes lie in the fact that it breaks the paradigm of `Comparisons` being leaf nodes and `Configurations` being branching nodes.
  - `AnyComparison` does not prepare its sub-comparisons, so we cannot inspect its members to find out which failed and which passed
  - `AnyComparison`'s nesting does not get revealed in places you might expect it to. (eg. `atef check`, `PreparedConfiguration.walk_comparisons`)  Phrased differently, AnyComparison sometimes hides its children from atef
  - Clicking the "next" button on an AnyComparison widget results in `tree_widget.currentItem()` to return None.  Despite the navigation being set up correctly and the item existing.  I added a clause to catch this but haven't found a better solution.
  - These complaints do not mean that `AnyComparison` should be removed.  It's just that in trying to patch these issues, I ran into situations where I was introducing brittle custom logic that could easily be broken
- I don't have better ideas at the moment, but "configuration" quickly reached the point of semantic saturation.  I would love to rename these fields to something more distinguishable, but that may require another migration.

</details>

[(current) example report output](https://github.com/pcdshub/atef/files/10504753/config_inspect_report.pdf)

## How Has This Been Tested?
Interactively

This PR, as well as much scratchwork and musings on [confluence](https://confluence.slac.stanford.edu/display/PCDS/atef+backend+details)


tangkong:
Lots of 
```
############################
    # Package would be ignored #
    ############################
    Python recognizes 'atef.tests' as an importable package,
    but it is not listed in the `packages` configuration of setuptools.
```
in the CI.  Need to decide if that's ok

tangkong:
pip 3.11 failed once on the ping-127.0.0.1 test, but passed on a re-run.  race condition somewhere?

klauer:
Hmm, it appears that I may have used a 

> 

Not sure what happened here - the migration tool should have used the latest cookiecutter, but it looks like it was slightly behind. Please fix the line in pyproject.toml to have "atef*" instead of just "atef":

https://github.com/pcdshub/cookiecutter-pcds-python/blob/15529c1e5d956843e6f2500de57cdc3900b6a07a/%7B%7B%20cookiecutter.folder_name%20%7D%7D/pyproject.toml#L27

klauer:
The other one:
> -   - package bluesky-queueserver-0.0.6-py37h89c1867_0 requires python_abi 3.7.* *_cp37m, but none of the providers can be installed

Is bluesky-queueserver _actually_ arch-specific or is it missing noarch in the recipe? https://github.com/conda-forge/bluesky-queueserver-feedstock/blob/main/recipe/meta.yaml

tangkong:
It's probably time to close this one out, it's gotten too fat and I want to start work on active checkouts

atef/config.py
```diff
@@ -1410,3 +1414,23 @@ def get_result_from_comparison(
         )
 
     return item, item.result
+
+
+async def run_passive_step(
```
ZLLentz, line 1419: Small thought: is this the right module to put this coroutine?

atef/widgets/config/page.py
```diff
@@ -1659,3 +1659,16 @@ def clean_up_any_comparison(self) -> None:
         - Cleans up all the sub-nodes.
         """
         self.tree_item.takeChildren()
+
+
+class ProcedureGroupPage(ConfigurationGroupPage):
```
ZLLentz, line 1664: Is this used at all yet?

atef/report.py
```diff
@@ -0,0 +1,546 @@
...
+    """
+    Document template to auto-gather table of contents and manage styles
+    Also holds config as an attempt to consolidate information, settings
+
+    Should be initialized with the report file path and configuration file
+    to render.
+
+    .. code-block:: python
+
+        doc = AtefReport('/path/to/pdf/location/report.pdf, config=file)
```
ZLLentz, line 51:
```suggestion
        doc = AtefReport('/path/to/pdf/location/report.pdf', config=file)
```

atef/widgets/config/run.py
```diff
@@ -0,0 +1,285 @@
...
+    # add RunCheck to end of layout
+    check_widget = RunCheck(configs=configs)
+    widget.layout().addWidget(check_widget)
+    widget.run_check = check_widget
+
+    return widget
+
+
+def disable_widget(widget: QWidget) -> QWidget:
+    """ Disable widget, recurse through layouts """
```
ZLLentz, line 85: qt learning moment for me: what does this do that a simple `widget.setEnabled(False)` doesn't do? Is it possible to have a disabled parent widget with enabled children?

atef/widgets/config/run.py
```diff
@@ -0,0 +1,285 @@
...
+        if next_widget:
+            self.setup_next_button(next_widget)
+
+        self.setup_verify_button()
+
+    def infer_step_type(self, config: Union[PreparedComparison, ProcedureStep]) -> str:
+        if hasattr(config, 'compare'):
+            return 'passive'
+        elif hasattr(config, 'run'):
+            return 'active'
```
ZLLentz, line 181: Is there a more direct way to do this?

atef/widgets/config/utils.py
```diff
@@ -463,6 +465,117 @@ def _open_multiline(self, to_select: Optional[List[str]] = None) -> None:
         self.edit_items(to_select, new_pvs)
 
 
+class Toggle(QCheckBox):
```
ZLLentz, line 468: Missing docstring, maybe just a brief "this is the widget that looks like this"

atef/widgets/config/utils.py
```diff
@@ -686,3 +799,67 @@ def cast_dataclass(data: Any, new_type: Type) -> Any:
         if key in field_names
     }
     return new_type(**new_kwargs)
+
+
+class MultiInputDialog(QtWidgets.QDialog):
```
ZLLentz, line 804: Missing brief docstring, what is this for?

ZLLentz (approved):
Small thoughts and questions from reading through the source
Next I'll try to open the gui- there aren't any test procedures, right? Just running the passive checks in the gui?

ZLLentz:
Usage thoughts/ideas for follow-ups:

I like the report that this generates a lot. It's very verbose but I think that's appropriate. I hope nobody prints these... the lfe.json demo generated an 116 page report.

There are some rendering oddities in the table of contents that are probably a bit more pronounced due to the way my config is structured. This makes the TOC a bit hard to parse since it's hard to differentiate the header levels.

image

The summary pages are especially good, but maybe something about the parameterization needs to make it into the step name since these all look the same:

image

For running in the gui, it'd be super neat if we could add a column on the left of the tree that just has the ?,X,check marks so it's easy to quickly scan down the side and see what failed

image

Last, running in the gui is pretty opaque, and I think some users will be confused if anything is happening at all between clicking "run" and waiting for the results on a big file. I wonder if there could also be a loading/running indicator somewhere.

atef/widgets/config/page.py
tangkong: Nope, This is probably left over from when my goal was to do active checkouts, before I pared down the scope of this PR.

atef/widgets/config/run.py
tangkong: Exactly.  For example, you might disable a parent widget, but the combo boxes or buttons inside a scroll area were still interactable

tangkong:
> Small thoughts and questions from reading through the source Next I'll try to open the gui- there aren't any test procedures, right? Just running the passive checks in the gui?

That's right, I haven't written any tests as of yet.  I've been testing against the all_fields.json config primarily, with running checks from the gui as the primary target

atef/widgets/config/utils.py
tangkong: I'll add that, this is for the input widget that pops up after setting the report file name, lets the user set the header/footer/version/author info.

atef/widgets/config/run.py
ZLLentz: That's wildly annoying behavior

atef/__init__.py
```diff
@@ -1,7 +1,3 @@
-from ._version import get_versions
-
-__version__ = get_versions()["version"]
-del get_versions
-
+from .version import __version__  # noqa: F401
```
klauer, line 1: Commenting here because image files can't be commented on:

1. I think we need to add `*.jpeg` to `MANIFEST.in` (or figure out the modern equivalent)
2. The SLAC logo is close to a megabyte, which seems kinda excessive. Unfortunately it's already deep your git history, so changing it is probably not worth the effort at this point...

atef/config.py
```diff
@@ -602,9 +606,10 @@ def walk_groups(
     ) -> Generator[PreparedGroup, None, None]:
         """Walk through the prepared groups."""
         for config in self.configs:
-            if isinstance(config, PreparedGroup):
+            if isinstance(config, get_args(AnyPreparedConfiguration)):
```
klauer, line 609: For reference, this becomes:
```python
isinstance(config, (
    PreparedDeviceConfiguration,
    PreparedGroup,
    PreparedPVConfiguration,
    PreparedToolConfiguration
   )
 )
```

I'm a bit confused why `walk_groups` turned into this: it's yielding things that aren't just `PreparedGroup`
I guess the annotation needs updating or the function needs renaming?

atef/config.py
klauer: Maybe we might be missing `atef.prepared` to separate out 'checks' vs 'configs' vs 'prepared configs' vs 'widgets/pages'?  That sounds like... too much work.

atef/procedure.py
```diff
@@ -1,13 +1,52 @@
...
+from atef.type_hints import AnyPath
+from atef.yaml_support import init_yaml_support
 
 from . import serialization
 
+run_engine_singleton = None
+db_singleton = None
+
+
+def incomplete_result():
```
klauer, line 31: This got defined here and in `atef.check`, it looks like?

atef/procedure.py
```diff
@@ -1,13 +1,52 @@
...
+    return incomplete
+
+
+def setup_temp_runengine():
+    global run_engine_singleton
```
klauer, line 36: Next iteration suggestion: maybe we can wrap this in a `BlueskyState` class or something instead of doing module-global stuff

atef/procedure.py
```diff
@@ -81,27 +159,142 @@ class DeviceConfiguration:
...
+
+
+@dataclass
+class PassiveStep(ProcedureStep):
+    """A step that runs a passive checkout file"""
+    filepath: pathlib.Path = field(default_factory=pathlib.Path)
+
+    def _run(self) -> Result:
+        """ Load, prepare, and run the passive step """
+        if self.filepath.suffix == '.json':
```
klauer, line 188: Not required in this PR, but a refactor thought/suggestion: `ConfigurationFile` should do the filename check and use the appropriate loader

-> `ConfigurationFile.from_filename(self.filepath)`

atef/procedure.py
```diff
@@ -81,27 +159,142 @@ class DeviceConfiguration:
...
+            config = ConfigurationFile.from_json(self.filepath)
+        else:
+            config = ConfigurationFile.from_yaml(self.filepath)
+
+        # prepare file
```
klauer, line 193:
```suggestion
```

atef/procedure.py
```diff
@@ -81,27 +159,142 @@ class DeviceConfiguration:
...
+
+        return result
 
 
 @dataclass
 class ProcedureGroup(ProcedureStep):
     """A group of procedure steps (or nested groups)."""
     #: Steps included in the procedure.
-    steps: Sequence[Union[ProcedureStep, ProcedureGroup]]
+    steps: Sequence[Union[ProcedureStep, ProcedureGroup]] = \
```
klauer, line 209: Is the `\` continuation really required with our flake8 settings? I'd think probably not?

atef/procedure.py
```diff
@@ -81,27 +159,142 @@ class DeviceConfiguration:
...
+        return self.root.compare()
+
+
+def run_procedure(
+    process: ProcedureFile,
+):
+    """
+    Run a procedure given a ProcedureFile.
+
+    Gross skeleton code I am organizing my thoughts
```
klauer, line 290: 😀

atef/report.py
```diff
@@ -0,0 +1,546 @@
...
+
+from atef.check import Result
+from atef.config import (PreparedComparison, PreparedConfiguration,
+                         PreparedDeviceConfiguration, PreparedFile,
+                         PreparedGroup, PreparedPVConfiguration,
+                         PreparedSignalComparison, PreparedToolComparison,
+                         PreparedToolConfiguration)
+from atef.enums import Severity
+
+h1 = PS(name='Heading1', fontSize=16, leading=20)
```
klauer, line 31: In the future, all these style things could all be part of another configuration file setting, I'd imagine

atef/report.py
```diff
@@ -0,0 +1,546 @@
...
+
+h2 = PS(name='Heading2', fontSize=14, leading=17)
+
+l0 = PS(name='list0', fontSize=12, leading=15, leftIndent=0,
+        rightIndent=0, spaceBefore=12, spaceAfter=0)
+
+styles = getSampleStyleSheet()
+
+
+class AtefReport(BaseDocTemplate):
```
klauer, line 41: The report is pretty awesome!

We have some tweaks to do, of course, but I'm not going to waste either of our time nitpicking on reportlab stuff. Its API doesn't make anything easy...

atef/report.py
```diff
@@ -0,0 +1,546 @@
...
+    def create_report(self) -> None:
+        """ Build the final report. """
+        raise NotImplementedError()
+
+
+class PassiveAtefReport(AtefReport):
+    """
+    Report for Passive Checkouts.  Assumes specific PreparedFile structure
+    """
+    __doc__ = __doc__ + AtefReport.__doc__
```
klauer, line 278:
```suggestion
    """ + AtefReport.__doc__
```

atef/report.py
```diff
@@ -0,0 +1,546 @@
...
+                style.append(['LINEABOVE', (0, i+1), (-1, i+1), 1, colors.lightgrey])
+
+        table = platypus.Table(
+            table_data, style=style
+        )
+
+        story.append(table)
+        story.append(platypus.PageBreak())
+
+    def walk_config_file(
```
klauer, line 347: Hmm, next refactor opportunity: this function doesn't seem to belong here. It could be out in a utility somewhere as it doesn't rely on `self`

atef/widgets/config/run.py
klauer: This may be something to review again in the future; maybe make a TODO?

klauer (commented):
This is a heck of a PR - I'm glad you limited the scope.

I still have a few more files left to check, but don't want to leave this review open.
I think this is a good check-in point and I really like the direction it's going. I'd be just fine with merging and iterating on it 👍

atef/procedure.py
tangkong: If I'm understanding my past self, I think I define the function here so I can pass into the dataclass default_factory, which doesn't like lambdas.  The shortcut `incomplete` was defined in `atef.check`, and just passed through here.  

I suppose I could redefine the helper function in each module I use it, maybe I took DRY a little too far.

atef/procedure.py
tangkong: It's actually improper syntax to split the line there without the `\`

atef/procedure.py
tangkong: I stand by this

atef/report.py
tangkong: If you do have thoughts for tweaks/nitpicks, let me know and I can make an issue.  I know I'll forget otherwise

atef/config.py
tangkong: The annotation definitely needs updating, good point.  

My rationale here is that this method wants to walk dataclasses that hold **groups** of comparisons.  In my mind, these 4 subclasses of `PreparedConfiguration` are siblings in that sense.  

Pragmatically, this method wasn't used anywhere else and I just mutated it to suit my needs :)

tangkong:
I tried removing the big file from the git history, which worked well locally (great even!)  In doing so I had rewritten the commit history via `git filter-branch`.  This stepped through every commit in atef's history... apparently.  And the diff became +25,000 / -2000.  I think there's probably a way to limit the point at which rewrites start, but I've been spooked. 

I force pushed myself back to the start.  Our poor bloated commit history.

atef/procedure.py
klauer: I meant that the line splitting itself might not have been required

atef/procedure.py
klauer: Yeah, the "no lambda" requirement makes sense. I was pointing at this:

```
$ git grep 'def incomplete_result'
atef/config.py:def incomplete_result():
atef/procedure.py:def incomplete_result():
```

But we've already spent more time than it's worth for removing 2 lines of code, so... yeah

atef/procedure.py
tangkong: aaaah right that's fair.  I'm starting to feel like I'm more strict than flake8 on line length...

atef/widgets/config/run.py
tangkong: I'll leave a TODO.  I wasn't immediately satisfied with just checking against step types, since as we add new steps we'd have to remember to maintain that list.  I'll circle back to this later

atef/__init__.py
tangkong: Added to the manifest and replaced with a smaller file.  Removing the logo from the commit history is something we definitely can do, but would be a bit disruptive to our commit history

tangkong:
I think I've either resolved or made issues for all of the comments.  Thanks for the reviews!  If there are no objections within the next day I'll go ahead and merge this and start addressing issues in separate PR's

klauer:
🥳

PR #142, opened 2023-01-26
BLD: simple dummy action

tangkong:
## Description
A minimal gha job (I think) to initialize actions for this repository

## Motivation and Context
We can't run actions in pull requests if gha has not been run in this repository before.  This should be removed in a future PR.

## How Has This Been Tested?
N/A

Here

### DIFF
--- a/.github/workflows/init_gha.yml
+++ b/.github/workflows/init_gha.yml
@@ -1,0 +1,8 @@
+name: GitHub Actions initialization
+run_name: ${{ github.actor }} is testing GitHub Actions
+on: [push]
+jobs:
+  dummy-job:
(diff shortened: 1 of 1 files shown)


klauer (approved):

tangkong:
Travis fails due to the same lightpath issues that have plagued recent builds (specifically #140).  With nothing changing in our ci helpers and source code, I suspect something having changed on Travis' side. (Test builds passed in our most recent commit history) 

New (to my knowledge) is a set of config warnings complaining about our use of Travis' `workspaces` feature, which is pretty much in line with their suggested usage...  Here's to hoping GHA does better.

PR #145, opened 2023-02-03
FIX: transposed the out and free run checks in demo config

ZLLentz:
## Description

Previously, when I was transferring the old demos to the new config format by hand, I made a mistake. This PR fixes the mistake.

## Motivation and Context

The demos should at least be well-formed.

## How Has This Been Tested?

Interactively

N/A

### DIFF
--- a/atef/tests/configs/lfe.json
+++ b/atef/tests/configs/lfe.json
@@ -145,23 +145,6 @@
             ],
             "detector.trigger_mode": [
               {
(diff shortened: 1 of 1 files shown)


klauer (approved):
Oops, good catch 👍

tangkong (approved):
LGTM.  May GHA deliver us from travis 🙏

PR #149, opened 2023-02-16
FIX: new file handling

ZLLentz:
## Description

Fix an issue where the "new file" path didn't run the same code as the "open file" path, so now the GUI works the same whether we start from an existing file or a new file.

## Motivation and Context

Pretty big collective blind spot- we've been testing given existing files rather than from blank.

## How Has This Been Tested?

Interactively

N/A

### DIFF
--- a/atef/widgets/config/window.py
+++ b/atef/widgets/config/window.py
@@ -119,8 +119,7 @@ class Window(DesignerDisplay, QMainWindow):
 
         The parameters are open as to accept inputs from any signal.
         """
-        widget = DualTree(config_file=ConfigurationFile())
-        self.tab_widget.addTab(widget, self.get_tab_name())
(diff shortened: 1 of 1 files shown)


atef/widgets/config/window.py
```diff
@@ -491,7 +507,7 @@ def __init__(
         self,
         *args,
         config_file: ConfigurationFile,
-        full_path: str,
+        full_path: Optional[str] = None,
```
ZLLentz, line 510: This was used in a previous version of the PR (the 5 minute version) but I could revert it now if it's distasteful. The widget works properly when full_path is given as None.

klauer (approved):

tangkong (approved):
LGTM.  happy to have this fixed

Issue #147, opened 2023-02-07
Collected refactor thoughts and ideas


tangkong:
## Description
A laundry list of ideas originating from #140 , which introduced a variety of opportunities for simplification and refactoring.  These may or may not be implemented, and likely warrant additional discussion

* ~we might be missing `atef.prepared` to separate out 'checks' vs 'configs' vs 'prepared configs' vs 'widgets/pages'?  That sounds like... too much work.~  It wasn't too much work.  #158 
* Perhaps wrapping the `RunEngine` instance in a `BlueskyState` class, rather than module-global manipulation
* `ConfigurationFile` could do the filename check and use appropriate loaders. 
  * `Configuration file.from_filename(self.filepath)`
* Move walk_config_file out of the report class.  Possibly move all the walking methods into a separate module
* Move report style settings into a configuration file setting
* Consolidate / properly classify all the `utils` in this package.  They have gotten populated enough to possibly justify new submodules.

## Possible Solution
Do the dirty work and clean up this package

## Your Environment
pcds-5.6.0

PR #150, opened 2023-03-07
ENH: add active checkout framework

tangkong:
## Description
- adds an option to open an active checkout at startup
- reworks `normalize_annotation` to use typing methods instead of parsing the string type hints
- Refactors the `atef.widgets.config` submodule (more on this below)
- gathers `.from_json()` and `.from_yaml()` into a `.from_filename()` classmethod
- Adds active checkout `ProcedureGroup` and a minimal `DescriptionStep` as a proof of concept
- Basically duplicates `ConfigurationGroupPage` into `ProcedureGroupPage`
  - There's surely opportunities for code reuse here, but I didn't want to get into it
- Adds verify functionality to RunCheck, combining the result of the step and result of the verification.
  - ProcedureStep has been augmented accordingly
- added a couple tests

Still outstanding
- Make sure run button behavior is good
- Reports for active checkouts (maybe another PR)
- ... the rest of the active checkout widgets (definitely another PR)

### `atef.widgets.config` Refactor
What was done:
- `data.py` -> `data_base.py`, `data_passive.py`, and `data_active.py`
  - `data.py` originally held all of the "data entry" widgets and helpers.  With the edit-run split, I wanted to make it clear where to find specific bits
  - `data_base.py` holds base classes useful to both the edit and run widgets (`DataWidget`, `NameDescTagsWidget`)
  - `data_passive.py` holds widgets used to edit passive checkouts (Comparisons, generally).  
  - `data_active.py` holds widgets used to edit active checkouts (currently just `DescriptionStep`)
    - This is where the old `atef.widgets.config.procedure` file got placed.  Much of this is not being used, but has been kept for now
- `run.py` -> `run_base.py`, `run_active.py`, `run_passive.py`
  - Similarly, run-mode widgets were split up.  `run_base` holds widgets like `RunCheck`, while `DescriptionStepRunWidget` lives in the `run_active.py` file

I think this helps clarify the relations between modules and possibly prevent circular imports.   the `*_base.py` files are at the lowest level and should not rely on any other `atef.widget.config` submodue.  Conversely, `atef.widgets.config.window` and `atef.widgets.config.page` are at the highest level, and can import from the `*_base`, `*_active`, and `*_passive` submodules.  

All `PageWidget`s should live in `atef.widgets.config.page`, and all `DataWidget`s live in the `*_{base, active, passive}` submodules

----
## Motivation and Context
In an effort to keep things review-able, I'm coming up for air here before going on.  I think I'm at a place where adding an additional step is well defined.  The process should be (I think):
- Identify which type of checkout your step applies to (active or passive)
- Create a data-widget for editing the settings for this step in the appropriate submodule (`data_passive` or `data_active`)
- Create a run-widget for displaying the result of the step in the appropriate submodule (`run_passive` or `run_active`)
  - If no run-widget is provided, atef will freeze the data widget and add a `RunCheck` widget
- Add references to your widget in `page.py`
  - if active, the edit widget should be mapped in `StepPage.step_map`, and the run widget mapped in `RunStepPage.run_widget_map`.  
  - if passive (adding a comparison), add your edit-widget to the `ComparisonPage.comp_map`
  - The edit widget should be mapped in `atef.widgets.config.page.PAGE_MAP`. 
  - Add your step to `{Configuration, Procedure}GroupPage.config_cls_options` so it can be created from the root page.

Currently there are no run-widgets for passive checkouts, meaning the `run_passive.py` file is empty.  Once some of these are created, we will have to reference them in `atef.widgets.config.window._edit_to_run_page`

Some of the refactor bits address #147 

----

## How Has This Been Tested?
A couple tests have been added.

This PR, and in some docstrings, though I should write up some docs too...

Selection after opening a new file
image

ProcedureGroupPage
image

DescriptionStep (edit page)
image

DescriptionStep (run page), before run/verify
image

After run (incomplete due to required Verify)
image

After run (failed verify)
image


tangkong:
I think I'm at a good stopping place feature-wise.  I'm going to spend tomorrow writing some tests (hopefully), and maybe I can start working on more actual active checkout steps / functionality by the end of this week...

tangkong:
py3.9 conda tests failed due to segfault but passed on a re-try.  🤷 qt-testing-things

atef/procedure.py
```diff
@@ -81,8 +83,43 @@ def run(self) -> Result:
...
 
-        self.result = result
-        return result
+        # stash step result
+        self.step_result = result
+        # return the overall result, including verification
+        return self.result
+
+    @property
+    def result(self) -> Result:
```
klauer, line 92: If this is something we want serialized (i.e., when saving procedure results/data), it might not belong in a property.

atef/procedure.py
```diff
@@ -252,6 +285,15 @@ def walk_steps(self) -> Generator[AnyProcedure, None, None]:
         yield self.root
         yield from self.root.walk_steps()
 
+    @classmethod
+    def from_filename(cls, filename: AnyPath) -> ProcedureFile:
+        path = pathlib.Path(filename)
+        if path.suffix == '.json':
```
klauer, line 291: Same thing here about case sensitivity

atef/qt_helpers.py
```diff
@@ -49,78 +49,70 @@ def my_class:
...
+        # 3. A list of values -> make a QDataclassList
+        # 4. A list of dataclasses -> QDataclassList (object)
+        origin = get_origin(type_hint)
+        args = get_args(type_hint)
+
+        if not origin:
+            # a raw type, no Union, Optional, etc
+            NestedClass = QDataclassValue
+            dtype = type_hint
+        elif origin == dict:
```
klauer, line 86: I would always choose `type_a is type_b` in these scenarios (instead of `==`). Not strictly necessary to change, though, as the end result is the same for these builtins.

atef/procedure.py
tangkong: I think I stash it under `_result` for that reason actually.  That should get updated every time the property gets called

atef/ui/general_procedure_widget.ui
```diff
@@ -0,0 +1,56 @@
...
+     </property>
+    </widget>
+   </item>
+   <item row="0" column="1">
+    <widget class="QComboBox" name="verify_combo"/>
+   </item>
+   <item row="1" column="0">
+    <widget class="QLabel" name="label">
+     <property name="text">
+      <string>Step Sucess Required: </string>
```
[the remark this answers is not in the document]
tangkong: Some typos really look glaring in the github diff.  

I struggled with wording here.  "Step must success for step to be marked as successful" 😵  

"Step must succeed" seems ok to me.

atef/widgets/config/window.py
```diff
@@ -88,6 +85,28 @@ def welcome_user(self):
         new_button.clicked.connect(self.new_file)
         welcome_box.exec()
 
+    def _passive_or_active(self) -> str:
```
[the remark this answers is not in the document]
tangkong: ooh or just a landing page / widget with those options.  avoids having a floating window to keep track of

atef/ui/run_check.ui
```diff
@@ -27,14 +27,21 @@
   </property>
```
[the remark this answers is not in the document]
tangkong: Agreed on setting up a spacer between the last widget and the RunCheck widget, I can do that here easily enough.  

I think I originally placed it in the bottom as to avoid conflicting with the name-description-tags widget when autogenerating a run-page.  In that context I like it there.  

Do you mean somewhat outside the page (outside the tab widget container)?  more in the window than the page?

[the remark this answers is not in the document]
klauer: I guess top or bottom, either would be fine as far as I'm concerned. 

A consistent spot where I can always expect to click 'next' without moving my mouse around is what I'd hope for.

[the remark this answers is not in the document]
tangkong: In my setup the next button is in a consistent place (once the RunCheck widget is spaced to the bottom), is it not for you?  Does it shift left/right?

atef/procedure.py
klauer: Could `_result` remain `result`, and then this be something like `overall_result` property or perhaps `get_display_result()` or `get_combined_result()` or something?

tangkong: That's right, making `result` a property backed by calculations was done to keep the interface between `Configuration` and `ProcedureStep` similar.  Both have a `result` that the `RunCheck` widget looks for

I was worried about future consumers grabbing an "overall result" that isn't updated with the most recent verify and step results.  I'd be satisfied with renaming `_result` -> `combined_result` and leaving property as is, but does that also jive with you?

atef/tests/test_widgets.py
```diff
@@ -161,21 +168,21 @@ def test_serialization(group):
...
-    qt_utils.save_widget_screenshot(widget, prefix=request.node.name)
-    widget.close()
-    widget.deleteLater()
+# @parametrized_groups
+# def test_create_widget(request: pytest.FixtureRequest, group: ProcedureStep):
+#     widget = procedure_step_to_widget(group)
+#     widget.show()
+#     qt_utils.save_widget_screenshot(widget, prefix=request.node.name)
+#     widget.close()
+#     widget.deleteLater()
```
ZLLentz, line 177: Maybe this should include a short comment about why the test is disabled via comment

atef/widgets/__init__.py
```diff
@@ -1,3 +1,2 @@
 # TODO: Configuration GUI needs to be updated for ConfigurationGroup
 # from .config import *  # NOQA
```
ZLLentz, line 2: I think these are old and can be removed
```suggestion
```

atef/widgets/config/run_active.py
```diff
@@ -0,0 +1,39 @@
+"""
+Widgets for config gui's active-checkout run mode.
+Widgets should mamp onto edit widgets from atef.widgets.config.data_active
```
ZLLentz, line 3:
```suggestion
Widgets should map onto edit widgets from atef.widgets.config.data_active
```

ZLLentz (approved):
The code looks good
I have a comment or two maybe on the threads Ken started here and then I want to clone and run this branch

atef/ui/run_check.ui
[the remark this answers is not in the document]
ZLLentz: Is this pulling us slowly in the direction of _always_ having some sort of prev/next navigation bar at the top or bottom rather than adding it to the page widgets as needed?

atef/widgets/config/run_base.py
```diff
@@ -0,0 +1,416 @@
...
+            return None
+
+    @property
+    def verify_results(self) -> List[Result]:
+        try:
+            return [c.verify_result for c in self.data]
+        except AttributeError:
+            return None
+
+    def setup_next_button(self, next_item=None) -> None:
```
ZLLentz, line 341: I think something is incomplete or buggy in the next button

image

atef/ui/run_check.ui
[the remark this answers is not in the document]
klauer: Not horizontally, just vertically in the spacer-less case right now. I think what I was after is that the "Page" can be pretty large and we may end up putting a scroll area there. In which case, I would not want to have to scroll to see the 'prev/next' buttons, but rather have it part of the parent layout. (This is all a little difficult to explain in words - and a bit forward-looking. I think it's perfectly fine to wait to address this and see how you think it works best with the remainder of the config widgets and all)

> Is this pulling us slowly in the direction of _always_ having some sort of prev/next navigation bar at the top or bottom rather than adding it to the page widgets as needed?

I like the idea of always having prev/next, verify/run/status/etc can be selectively shown depending on the mode

atef/widgets/config/run_base.py
```diff
@@ -0,0 +1,416 @@
...
+        icon = self.style().standardIcon(self.style_icons[Severity.warning])
+        self.result_label.setPixmap(icon.pixmap(25, 25))
+        self.data = data
+
+        if data:
+            self.setup_buttons(configs=data)
+            # initialize tooltip
+            self.update_status()
+
+    def setup_buttons(self, configs, next_widget: AtefItem = None) -> None:
```
ZLLentz, line 230: Somehow this version of the buttons is bleeding over into the passive checks, is this expected?

image

atef/widgets/config/run_base.py
```diff
@@ -0,0 +1,416 @@
...
+
+        # step result
+        if self.step_results:
+            self.update_icon(self.run_success_label, self.step_results)
+            self.update_label_tooltip(self.run_success_label, self.step_results)
+
+            self.update_icon(self.verify_label, self.verify_results)
+            self.update_label_tooltip(self.verify_label, self.verify_results)
+
+    def update_icon(self, label: QLabel, results: List[Result]) -> None:
```
ZLLentz, line 284: I'm getting a green checkmark in the left slot before running any procedures, is this expected?

image

atef/widgets/config/run_base.py
tangkong: hm, it is not.  I think this has to do with AnyComparison being weird.  I modify the RunCheck widget based on the config associated with the widget, and I can't find a way to get the proper configs/comparisons for AnyComparison (or haven't found a clean way, at least).

I'll look into ir more

ZLLentz:
OK I clicked around a bunch (starting from a blank GUI and sticking with the implemented Description Step only) and I think the user experience is pretty reasonable :+1: I pointed out any oddities I saw.

atef/widgets/config/run_base.py
tangkong: Yep, if you don't require the step to be verified or succeed, then it passes.  You have to set one of those to true for it to be a "?". 

As you mention it, it's probably good to set both to default to True

ZLLentz: I agree, it seems a bit odd to by default allow every procedure to fail

atef/widgets/config/run_base.py
ZLLentz: :fire: AnyComparison :fire: strikes again

atef/widgets/config/run_base.py
tangkong: I figured out why this happened about an hour ago and only just managed to fix it.  Qt being rascally as usual

atef/widgets/config/run_active.py
tangkong: 🤦

atef/procedure.py
tangkong: Renamed `_result` -> `combined_result` in [19287a7](https://github.com/pcdshub/atef/pull/150/commits/19287a72a182cc77520f34f2d58a4b8f49f1e901)

tangkong:
I think I've addressed the less involved issues mentioned above, thanks for the reviews.  If there are no further requests by next Tuesday when I return, I'll merge and continue on.  

As always, an atef PR spawns more issues than it closes

ZLLentz:
`atef` issues are the hydra, confirmed

PR #155, opened 2023-03-15
ENH: Add passive step to active checkout

tangkong:
## Description
Adds the passive-step to active checkouts.  That sentence is confusing, so phrased another way: Adds the ability to run passive checkouts as a part of an active checkout.

- commits to (re)implementing the model-tree-view structure for the run step widgets
  - `ConfigTreeModel` can constructed with `TreeItem` be passed into a `QTreeView`
- Creates `PassiveEditWidget` and `PassiveRunWidget` for insertion into `StepPage` and `RunStepPage` respectively
- misc touchups I can no longer remember

## Motivation and Context
The first of many steps to be added to the active checkout toolbox. 

Refactors things a touch along the way, but I'm starting to think that will never stop.  

### Growing a tree
Prior to this PR the construction of a file's tree was distributed.  An `AtefItem` would be created when `EditTree.assemble_tree()` is called, and additional leaves are added as they are created.  This PR creates utilities that can create the whole tree without also creating pages and widgets.  This does, however, fragment the logic for tree creation (and by extension config file traversal).  

This is fragile, but in my opinion necessary.  I will soon write tests that make sure this logic stays in sync... hopefully

### Storage of results from the passive step
Currently the results of the passive step are distilled into a single `step_result` field.  There were iterations of this where the storage of the PreparedFile was considered.  Those attempts quickly fell apart due to `PreparedFile`'s and `PreparedConfig`'s being unserializable (they have locks and such)

A knock-on effect of this is that the status icons shown in the tree view and RunCheck widget are technically reading from different `Result` objects.  They are generated from the same ConfigurationFile, but I cannot rule out the possibility that the data they gather will become desynced at some point. 

## How Has This Been Tested?
Interactively, I should write tests.  The test I wrote is very finicky.  Well, qt tests are finicky in general.  locally the suite will sometimes segfault when running the whole suite together, but succeed if `test_widgets` is run in isolation.  Also switching `qtbot.add_widget` to `qtbot.addWidget` reduced the segfault frequency, even though they should be equivalent...  I'm not sure if it's worth keeping this test in if it fails so frequently (suite took 3(?) retries to pass on GHA, with 3.10 conda never passing)

Docstrings, though I should solidify docs on how to add steps.

image

image


klauer:
This PR and the status icon screenshot make me feel like atef is really (finally?) coming together, for some reason.

Nice work, @tangkong  😁

tangkong:
I sort of want to forge ahead here, now that I've solidified the steps needed to add a step/widget.  

I think I'll leave testing for a later effort, as qt segfaults have stumped me and I could spend countless hours generating / debugging tests.  If others have strong feelings about getting tests alongside this PR I can take the time out and work on them.  I just want to push toward the bluesky plans asap, since I think they'll be troublesome

ZLLentz:
I think it's fine to forge ahead. I have also been frustrated by getting qt tests to run stably... You almost want to start a full new process with each test else you're guaranteed to have some sort of leak.

atef/widgets/config/page.py
```diff
@@ -1470,6 +1472,41 @@ def table_drop_event(self, event: QDropEvent) -> None:
...
+            return
+
+        step_row = ComparisonRowWidget(data=new_step)
+        self.setup_row_buttons(
+            row_widget=step_row,
+            item=comp_item,
+            table=self.procedure_table,
+        )
+        self.procedure_table.setCellWidget(found_row, 0, step_row)
+        # self.update_combo_attrs()
```
ZLLentz, line 1508:
```suggestion
```

ZLLentz (commented):
I have friday brain so there might be more but this is the only think I saw in the code
I'll give this a spin

atef/widgets/config/run_active.py
ZLLentz, line 1: The purpose of this is so that existing passive configs + other groups of checks can be all be done at once as part of a larger config, right? Not to e.g. check a bunch of values after some other active steps, which will be taken care of in more specific active steps instead?

I'm trying to reorient my brain a bit

tangkong: That is correct, this was meant for the specific case where a passive checkout is run as a part of an active checkout.  I do envision other active steps checking values on their own, though I suppose if someone wanted to they could create a separate passive checkout to run after each of their active steps

I haven't considered nesting an active checkout within an active checkout using this ui interface.  I thought that nested ProcedureGroups would be enough to enable that kind of organization

atef/widgets/config/window.py
ZLLentz, line 1: I'm having some strange behavior. I've configured an active checkout to contain 2 description steps and 2 different passive steps.

1. The summaries in the active config don't update when the other files are updated (though I'm not sure they should). Maybe this just needs a refresh/reload file button?
2. When I go into active mode from a fresh file open, I get the check marks at the bottom that indicate that my tests have already been run and passed, despite me not ever clicking "Run"
image
3. When I run the full active checkout and then check the passive step pages, the internal model view looks like it hasn't been run. The bottom area says the test was run, but it had this check before anything was run as per oddity 2
image

atef/widgets/config/run_active.py
ZLLentz: I wasn't considering nested active checkouts either

atef/widgets/config/window.py
tangkong: 1: a good point, there's no monitoring of the loaded files.  A workaround is to just re-load the passive checkout in edit-mode, but a reload/refresh button would be nice
2: ???  I can't actually reproduce this, though if I create a new file -> click run -> edit -> return to run mode, the check marks aren't cleared and seem to be weird in general...  I will investigate
3: This has to do with needing to duplicate the PreparedFile I mentioned in the PR description text.  The RunCheck widget (and its status icons) all hold onto the same PreparedFile, but the tree view holds onto a copy of that.   Thinking about it again, I can probably fix this pretty easily...

ZLLentz: > I can't actually reproduce this

oh no

I can send the silly test files I made if it helps

tangkong: I think you must of run the step previously, your active config already has the `step_result` saved as successful
```
{'root': {'combined_result': {'reason': '', 'severity': <Severity.success: 0>},
          'description': None,
          'name': 'root',
          'parent': None,
          'step_result': {'reason': None, 'severity': <Severity.success: 0>},
...
```

This does beg the question of how to reset those results, they certainly shouldn't carry over when we go back and forth between edit and run mode.

ZLLentz: I did not realize that the step_result was saved, I did not expect that behavior.

tangkong: I think ultimately my responses to these issues have been addressed as follows (Thanks for the fresh eyes on the UX Zach)

1: In the interest of UI simplicity I left the buttons as they were, and added a "last loaded" text label indicator.  
2: **I set up the results to be cleared whenever the run-mode is enabled, and added a "clear_results" utility action in the menu.  I think the proper behavior is that if a user loads a checkout file, they don't want old results being carried over; the results should be saved in a report.**  While with passive checkouts this wasn't consciously considered due to the PreparedFiles being separate from the checkout file, active checkouts don't inherently have this separation.
3: I determined it was too difficult/cumbersome to link the transient PreparedFile created in the `.run()` method to the one held by the tree view, so I just added a "refresh status" button

klauer: I'd have expected the result to be cleared _if_ my configuration changes, not just switching from config/run mode. Still, clearing on every switch to run mode is preferable to not clearing at all.

I think it's worth us taking some time and talking about that combination of checkout+result in the dataclass.

A configuration is a long-term and reusable thing, but results are ephemeral and always subject to change. I always envisioned the results to be in separate structures, though that does make linking the two together more complicated.

(Regardless of the decision we make, I think it's fine to merge this as-is and revisit.)

atef/procedure.py
```diff
@@ -44,6 +44,12 @@ def __init__(self):
         self.RE.subscribe(self.db.insert)
 
 
+def walk_steps(step: ProcedureStep) -> Generator[ProcedureStep, None, None]:
```
klauer, line 47: A few functions in this PR could use docstrings. I know they're simple methods in general, but let's just take care of it now before we fall behind.

atef/widgets/config/run_base.py
```diff
@@ -423,3 +399,43 @@ def __init__(self, *args, **kwargs):
...
+            if prepared_file:
+                # Grab relevant comps/configs so tree item can hold results
+                prep_configs = get_relevant_configs_comps(prepared_file, cfg)
+            else:
+                prep_configs = None
+            item = TreeItem(cfg, prepared_data=prep_configs)
+            create_tree_items(cfg, item, prepared_file=prepared_file)
+            parent.addChild(item)
+    except Exception as ex:
+        logger.debug(ex)
```
klauer, line 424: This try/except block looks like it needs some review.

(`except Exception` with a large block of code is making me think of my scary [pmgr](https://github.com/pcdshub/pmgr/blob/master/pmgr/CfgModel.py) (`pmgr$ git grep 'except Exception' |wc -l ->     156`) migration PR 😬 - even though this is nowhere near that level of course!)

atef/widgets/config/utils.py
```diff
@@ -900,3 +903,193 @@ def get_info(self) -> Dict[str, Any]:
...
+        return parent_item.childCount()
+
+    def columnCount(self, parent: QtCore.QModelIndex):
+        if not parent.isValid():
+            parent_item = self.root_item
+        else:
+            parent_item = parent.internalPointer()
+        return parent_item.columnCount()
+
+    def data(self, index, role):
```
klauer, line 995: Annotations?

atef/widgets/config/window.py
```diff
@@ -296,6 +301,40 @@ def print_report(self, *args, **kwargs):
...
+                'Confirm deletion',
+                (
+                    'Are you sure you want to clear the results of the checkout: '
+                    f'{config_file.root.name}?'
+                ),
+            )
+            if reply != QMessageBox.Yes:
+                return
+
+            if isinstance(config_file, ProcedureFile):
```
klauer, line 322: Nitpick: seems like this should be in a `clear_results` helper function outside of the GUI code?

klauer (approved):

atef/widgets/config/utils.py
tangkong: Apparently I was all over the place... This method's missing annotations and the one above it is missing a docstring... 🤦

atef/widgets/config/window.py
tangkong: done in e2d4988

atef/procedure.py
tangkong: 100%, I neglected to do my typical run-through after the methods/functions mostly crystalized.  [ae6bf9e](https://github.com/pcdshub/atef/pull/155/commits/ae6bf9e7a5f55587a057ea21b338215c1a97496a)

atef/widgets/config/window.py
tangkong: The run-mode results-clear only activates when a new RunTree is created, which in turn only happens if a new config is detected.  (So I _think_ the behavior you want is present)

By also clearing the results every time a new RunTree is created, we effectively never see the old results stored in the config json.  This is the right behavior, but as you mention, we should think about the separation of Results from their configs.  

Playing devil's advocate, I can see the bundled data-result (status quo) as providing an alternate way of saving the results.  A backup / redundant report, possibly.  (This is perhaps nice now, while the active checkout reports don't exist).

I'll continue to mull over this.

ZLLentz (approved):
No more thoughts, should be gtg

PR #158, opened 2023-03-28
MNT: refactor active checkouts to have Prepared variants

tangkong:
## Description
- Refactors the `Result` fields and and `.run()` methods for active checkout dataclasses into `Prepared*` procedure step dataclasses
- Moves the `Result` dataclass and related helpers into its own submodule.  Refactors accordingly
- Updates window and RunTree logic to reflect these changes
- Adds timestamp field to `Result` dataclass.  This field does not need to be specified at instantiation, and will be auto-filled.  This will also be ignored during comparisons between `Result` objects

## Motivation and Context
After some slack conversations, we decided that it was best to keep any transient information separated from the checkout specification.  This makes active checkouts mirror the structure of passive checkouts.  
It is expected that there will be information worth holding onto while running the checkout.  The first case of this is obviously the `Result` object, but as more steps are added this list may expand (sub-step results, images/metadata, etc)

## How Has This Been Tested?
Added a quick smoke test, I should add more

Docstrings, this PR

### DIFF
--- a/atef/tests/conftest.py
+++ b/atef/tests/conftest.py
@@ -1,7 +1,7 @@
 import contextlib
 import datetime
 import pathlib
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 import pydm
 import pydm.exception
@@ -12,6 +12,23 @@ from ..archive_device import ArchivedValue, ArchiverHelper
 
 TEST_PATH = pathlib.Path(__file__).parent.resolve()
 CONFIG_PATH = TEST_PATH / "configs"
+
+
+def passive_checkout_configs() -> List[pathlib.Path]:
+    filenames = ['lfe.json', 'all_fields.json', 'ping_localhost.json']
+    config_paths = [CONFIG_PATH / fn for fn in filenames]
+    return config_paths
+
+
+def active_checkout_configs() -> List[pathlib.Path]:
+    filenames = ['active_test.json']
+    config_paths = [CONFIG_PATH / fn for fn in filenames]
+    return config_paths
+
+
+PASSIVE_CONFIG_PATHS = passive_checkout_configs()
+ACTIVE_CONFIG_PATHS = active_checkout_configs()
+ALL_CONFIG_PATHS = PASSIVE_CONFIG_PATHS + ACTIVE_CONFIG_PATHS
 
 
 class MockEpicsArch:
--- a/atef/tests/conftest.py
+++ b/atef/tests/conftest.py
@@ -31,6 +31,21 @@ ACTIVE_CONFIG_PATHS = active_checkout_configs()
 ALL_CONFIG_PATHS = PASSIVE_CONFIG_PATHS + ACTIVE_CONFIG_PATHS
 
 
+@pytest.fixture(params=PASSIVE_CONFIG_PATHS)
+def passive_config_path(request) -> pathlib.Path:
+    return request.param
+
+
+@pytest.fixture(params=ACTIVE_CONFIG_PATHS)
+def active_config_path(request) -> pathlib.Path:
+    return request.param
+
+
+@pytest.fixture(params=ALL_CONFIG_PATHS)
+def all_config_path(request) -> pathlib.Path:
+    return request.param
+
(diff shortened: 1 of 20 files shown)


tangkong:
I'll stop this PR and ask for reviews before the diff gets too messy

ZLLentz (commented):
I think conceptually this is really good and I didn't notice anything wrong or out of place. I'm having trouble staying focused so I don't feel like I've given it as thorough a review as I would have liked, despite my best efforts.

atef/tests/test_procedure.py
```diff
@@ -40,22 +44,33 @@ def test_procedure_step_results(
...
+    assert prep_desc_step.step_result == pass_result
+
+
+@pytest.mark.parametrize('active_config', ACTIVE_CONFIG_PATHS)
+def test_prepared_procedure(active_config):
+    procedure_file = ProcedureFile.from_filename(filename=active_config)
+    # monkeypatch passive step to have
+    # simple smoke test
+    ppf = PreparedProcedureFile.from_origin(file=procedure_file)
+    asyncio.run(ppf.run())
```
klauer, line 76: With a `@pytest.mark.asyncio`, you can convert this test case into an async function, by the way:
```python
@pytest.mark.asyncio
async def test_prepared_procedure(active_config):
    ...
    await ppf.run()
```

atef/tests/conftest.py
```diff
@@ -14,6 +14,23 @@
...
+
+def active_checkout_configs() -> List[pathlib.Path]:
+    filenames = ['active_test.json']
+    config_paths = [CONFIG_PATH / fn for fn in filenames]
+    return config_paths
+
+
+PASSIVE_CONFIG_PATHS = passive_checkout_configs()
+ACTIVE_CONFIG_PATHS = active_checkout_configs()
+ALL_CONFIG_PATHS = PASSIVE_CONFIG_PATHS + ACTIVE_CONFIG_PATHS
```
klauer, line 31: For starters - this works just fine and doesn't need changing.

I wanted to point out that it can be nice to just make fixtures like so:
```python
@pytest.fixture(params=PASSIVE_CONFIG_PATHS)
def passive_config_path(request) -> pathlib.Path:
    return request.param
```

Then you just use `passive_config_path: pathlib.Path` as a test function argument and you don't have to repeat `pytest.mark.parametrize`

atef/widgets/config/run_base.py
```diff
@@ -169,6 +167,38 @@ def get_relevant_configs_comps(
...
+    -------
+    List[Union[PreparedProcedureStep, PreparedProcedureGroup]]
+        the PreparedProcedureSteps related to ``step``
+    """
+    # As of the writing of this docstring, this helper is only expected to return
+    # lists of length 1.  However in order to match the passive checkout workflow,
+    # we still return a list of relevant steps.
+    matched_steps = []
+    for pstep in walk_steps(prepared_file.root):
+        if pstep.origin == step:
```
klauer, line 197: Hmm, do we want an equality check here or an `is` check?

Dataclass equality checks implicitly do the [following](https://docs.python.org/3/library/dataclasses.html):
> eq: If true (the default), an __eq__() method will be generated. This method compares the class as if it were a tuple of its fields, in order. Both instances in the comparison must be of the identical type.

Whereas an `is` check would add the prepared step only if its origin were the exact object.

klauer (approved):
(Sorry, this review got buried in my queue somehow...)

I think this looks good and agree with the approach you're taking toward reorganization/refactoring here.

atef/tests/test_procedure.py
tangkong: Oh neato.  pytest's mysteries unravel itself slowly.

atef/tests/conftest.py
tangkong: pytest is a mysterious beast

tangkong:
I totally forgot about this until accidentally clicking the "Pull Requests" tab of my notifs @ZLLentz do you want a chance to re-review?  Otherwise I'll just merge and revisit issues no worries

Issue #152, opened 2023-03-09
GUI: OS-specific button layouts


tangkong:
## Expected Behavior
Buttons for windows should appear in the same place regardless of OS.  

Minimally, buttons should appear sensibly in all OS's even if they're in different places

## Current Behavior
The new checkout welcome screen splash does not show up the same in MacOS. Most offensively, the close button is not colored/emphasized.  The buttons still work as expected.

<img width="1134" alt="image" src="https://user-images.githubusercontent.com/35379409/224173577-d3ab80fa-05ea-4d72-b1dc-75de4a0840ce.png">

Are there more differences?  Perhaps... 🤔 

## Possible Solution
Possibly look at the button roles?

## Context
Looking at developing locally and hit upon this.

## Your Environment
pcds-5.6.0, and my own local environment

PR #159, opened 2023-04-07
ENH: add a Set-Value active checkout step

### LINKED ISSUE #39: Bug: Config GUI deallocation crash (wrapped object deleted)
## Expected Behavior

No crashes

## Current Behavior

Bug:
```python
(atef) PC98125:atef klauer$ atef config
  File "/Users/klauer/Repos/atef/atef/bin/config.py", line 574, in show_selected_display
    widget = item.get_widget()
  File "/Users/klauer/Repos/atef/atef/bin/config.py", line 620, in get_widget
    self.widget_cached = self.widget_class(*self.widget_args)
  File "/Users/klauer/Repos/atef/atef/bin/config.py", line 1231, in __init__
    self.initialize_idcomp()
  File "/Users/klauer/Repos/atef/atef/bin/config.py", line 1235, in initialize_idcomp
    self.initialize_config_name()
  File "/Users/klauer/Repos/atef/atef/bin/config.py", line 817, in initialize_config_name
    self.bridge.name.changed_value.connect(self.name_edit.setText)
RuntimeError: wrapped C/C++ object of type QDataclassValueForstr has been deleted
```

I was clicking around, alt-tabbed back to the PR, and saw the crash above

_Originally posted by @klauer in https://github.com/pcdshub/atef/pull/28#discussion_r853304100_

### LINKED ISSUE #143: ENH: Loading cursor
## Expected Behavior
It might be nice to have a better visual indicator that processes are running. 

## Current Behavior
The GUI just kinda pauses for a bit

## Possible Solution
On a widget level, can set and unset
``` python
from qtpy.QtCore import Qt
self.setCursor(Qt.WaitCursor)
self.unsetCursor()
```
(if you move outside the widget where the cursor was set, it will change back)

Can also set this at an application level, which would keep things consistent but then we'd have to grab the qapp

Would probably set up for
- Creating new run-tree (setting up widgets and new data structure)
- building report?

## Context
Some things are starting to take a bit longer and may be concerning

## Your Environment
pcds-5.6.0

tangkong:
## Description
Adds a step called `SetValueStep`, for setting values to signals/components, then running a set of comparisons.  Actions are stored as a list of target-value dataclasses, and checks are stored as a list of target-comparison dataclasses.  Targets can be specified as either a device - attr combination or raw PV string.   

Along the way:
- adds `TableWIdgetWithAddRow`, a subclass of QTableWidget that includes a `AddRowWidget`. This add row contains a button for adding rows of a specified widget.  (for better space efficiency)
image

- gets overly fancy with signal/component selectors
image  image

- adds support for ProcedureSteps that also contains checks, I imagine this will become more common in the future
- Fixes type hint parsing in QDataClassBridge acfe193
- adds a busy cursor decorator and puts it in a few places.  Closes #143 

- Attempts to disconnect some slots on widget destruction (addressing a crash identical to #39 but from a different source)  Perhaps closes #39 

## Motivation and Context

I wanted a basic action that can actually set PVs, at a very fine grained level.  I imagine this could serve as a fallback if other steps don't fit.

Notably missing (perhaps for followups, this is getting long)
- Option to use tools as checks (currently only ping?)
- look at drag and drop functionality for reordering TableWidgetWithAddRow
- keep tool menu open after component selection window, a QOL thing that I haven't figured out

## How Has This Been Tested?

interactively, and making sure the test suite passes.  Tested locally against the classic caproto random-walk IOC.
Existing test suite loads up the test config and toggles between edit and run, which makes sure the widgets all initialize

Docstrings, partially so far

image
image

### DIFF
--- a/atef/procedure.py
+++ b/atef/procedure.py
@@ -152,7 +152,7 @@ class Target:
     #: EPICS PV
     pv: Optional[str] = None
 
-    def to_signal(self) -> ophyd.EpicsSignal:
+    def to_signal(self) -> Optional[ophyd.EpicsSignal]:
         """
         Return the signal described by this Target.  First attempts to use the
         device + attr information to look up the signal in happi, falling back
@@ -186,7 +186,7 @@ class ValueToTarget(Target):
     value: Optional[PrimitiveType] = None
 
     # ophyd.Signal.set() parameters
-    #: connection timeout
+    #: write timeout
     timeout: Optional[float] = None
     #: settle time
     settle_time: Optional[float] = None
@@ -672,8 +672,7 @@ class PreparedSetValueStep(PreparedProcedureStep):
 
     def walk_comparisons(self) -> Generator[PreparedComparison, None, None]:
         """ Yields PreparedComparisons in this ProcedureStep """
-        for prep_comp in self.prepared_criteria:
-            yield prep_comp
+        yield from self.prepared_criteria
 
     async def _run(self) -> Result:
         """
@@ -850,7 +849,7 @@ class PreparedValueToSignal:
 
         pvts = cls(
             name=origin.name,
-            signal=origin.to_signal(),
+            signal=signal,
             value=origin.value,
             origin=origin,
         )
--- a/atef/procedure.py
+++ b/atef/procedure.py
@@ -21,7 +21,7 @@ import logging
 import pathlib
 from dataclasses import dataclass, field
 from typing import (Any, Dict, Generator, List, Literal, Optional, Sequence,
-                    Tuple, Union, cast)
+                    Union, cast)
 
 import apischema
 import databroker
@@ -129,11 +129,13 @@ class PassiveStep(ProcedureStep):
 @dataclass
 class SetValueStep(ProcedureStep):
     """ A step that sets one or more values and checks one or more values after """
-    actions: List[Tuple[Target, PrimitiveType]] = field(default_factory=list)
-    success_criteria: List[Tuple[Target, Comparison]] = field(default_factory=list)
+    actions: List[ValueToTarget] = field(default_factory=list)
+    success_criteria: List[ComparisonToTarget] = field(default_factory=list)
 
-    continue_on_fail: bool
-    require_action_success: bool
+    #: Continue performing actions if one fails
+    continue_on_fail: bool = False
+    #: Only mark the step_result as successful if all actions have succeeded
+    require_action_success: bool = True
 
 
 @dataclass
@@ -141,22 +143,37 @@ class Target:
     """
     A destination for a value.  Either an ophyd device+attr pair or EPICS PV
     """
+    #: name of target
+    name: Optional[str] = None
     #: device name and attr
-    device: Optional[str]
-    attr: Optional[str]
+    device: Optional[str] = None
+    attr: Optional[str] = None
     #: EPICS PV
-    pv: Optional[str]
+    pv: Optional[str] = None
 
     def to_signal(self) -> ophyd.EpicsSignal:
+        """ attempt to use happi first, failing that use stored PV info"""
         if self.device and self.attr:
             device = util.get_happi_device_by_name(self.device)
             signal = getattr(device, self.attr)
         elif self.pv:
             signal = ophyd.EpicsSignal(self.pv)
         else:
-            raise ValueError('No signal or device found')
+            return
 
         return signal
+
+
+@dataclass
+class ValueToTarget(Target):
+    #: the value to set to the target
+    value: Optional[PrimitiveType] = None
+
+
+@dataclass
+class ComparisonToTarget(Target):
+    #: the comparison to apply to the target
+    comparison: Optional[Comparison] = None
 
 
 @dataclass
@@ -595,7 +612,7 @@ class PreparedPassiveStep(PreparedProcedureStep):
 
 @dataclass
 class PreparedSetValueStep(PreparedProcedureStep):
-    prepared_actions: List[Tuple[ophyd.Signal, PrimitiveType]] = field(
+    prepared_actions: List[PreparedValueToSignal] = field(
         default_factory=list
     )
     prepared_criteria: Optional[List[PreparedSignalComparison]] = field(
@@ -618,14 +635,27 @@ class PreparedSetValueStep(PreparedProcedureStep):
             name=step.name
         )
 
-        for target, value in step.actions:
-            signal = target.to_signal()
-            prep_step.prepared_actions.append((signal, value))
+        for value_to_target in step.actions:
+            signal = value_to_target.to_signal()
+            value = value_to_target.value
+            prep_value_to_signal = PreparedValueToSignal(
+                signal=signal, value=value
+            )
+            prep_step.prepared_actions.append(prep_value_to_signal)
 
-        for target, comp in step.success_criteria:
-            signal = target.to_signal()
-            prep_comp = PreparedSignalComparison.from_pvname(pvname=signal.pvname,
-                                                             comparison=comp)
+        for comp_to_target in step.success_criteria:
+            signal = comp_to_target.to_signal()
+            comp = comp_to_target.comparison
+            prep_comp = PreparedSignalComparison.from_pvname(
+                pvname=signal.pvname, comparison=comp
+            )
             prep_step.prepared_criteria.append(prep_comp)
 
         return prep_step
+
+
+@dataclass
+class PreparedValueToSignal:
+    signal: ophyd.Signal
+    value: PrimitiveType
+    result: Result = field(default_factory=incomplete_result)
--- a/atef/procedure.py
+++ b/atef/procedure.py
@@ -770,8 +770,8 @@ class PreparedSetValueStep(PreparedProcedureStep):
             signal = comp_to_target.to_signal()
             comp = comp_to_target.comparison
             try:
-                prep_comp = PreparedSignalComparison.from_pvname(
-                    pvname=signal.pvname, comparison=comp
+                prep_comp = PreparedSignalComparison.from_signal(
+                    signal=signal, comparison=comp
                 )
                 prep_step.prepared_criteria.append(prep_comp)
             except Exception as ex:
--- a/atef/procedure.py
+++ b/atef/procedure.py
@@ -37,7 +37,7 @@ from atef.config import (ConfigurationFile, PreparedComparison, PreparedFile,
 from atef.enums import GroupResultMode, Severity
 from atef.exceptions import PreparedComparisonException
 from atef.result import Result, _summarize_result_severity, incomplete_result
-from atef.type_hints import AnyPath, PrimitiveType
+from atef.type_hints import AnyPath, Number, PrimitiveType
 from atef.yaml_support import init_yaml_support
 
 from . import serialization
@@ -193,9 +193,9 @@ class ValueToTarget(Target):
 
     # ophyd.Signal.set() parameters
     #: write timeout
-    timeout: Optional[float] = None
+    timeout: Optional[Number] = None
     #: settle time
-    settle_time: Optional[float] = None
+    settle_time: Optional[Number] = None
 
 
 @dataclass
--- a/atef/procedure.py
+++ b/atef/procedure.py
@@ -133,8 +133,8 @@ class SetValueStep(ProcedureStep):
     actions: List[ValueToTarget] = field(default_factory=list)
     success_criteria: List[ComparisonToTarget] = field(default_factory=list)
 
-    #: Continue performing actions if one fails
-    continue_on_fail: bool = False
+    #: Stop performing actions if one fails
+    halt_on_fail: bool = True
     #: Only mark the step_result as successful if all actions have succeeded
     require_action_success: bool = True
 
@@ -656,13 +656,13 @@ class PreparedSetValueStep(PreparedProcedureStep):
         self = cast(SetValueStep, self)
         for prep_action in self.prepared_actions:
             await prep_action.run()
-            if (not self.origin.continue_on_fail
+            if (self.origin.halt_on_fail
                     and prep_action.result.severity > Severity.success):
                 self.step_result = Result(
                     severity=Severity.error,
                     reason=f'action failed ({prep_action.name}), step halted'
                 )
-                break
+                return self.step_result
         for prep_criteria in self.prepared_criteria:
             await prep_criteria.compare()
 
--- a/atef/procedure.py
+++ b/atef/procedure.py
@@ -176,6 +176,12 @@ class ValueToTarget(Target):
     #: the value to set to the target
     value: Optional[PrimitiveType] = None
 
+    # ophyd.Signal.set() parameters
+    #: connection timeout
+    timeout: Optional[float] = None
+    #: settle time
+    settle_time: Optional[float] = None
+
 
 @dataclass
 class ComparisonToTarget(Target):
@@ -703,11 +709,9 @@ class PreparedSetValueStep(PreparedProcedureStep):
         )
 
         for value_to_target in step.actions:
-            signal = value_to_target.to_signal()
-            value = value_to_target.value
             try:
-                prep_value_to_signal = PreparedValueToSignal(
-                    signal=signal, value=value, name=value_to_target.name
+                prep_value_to_signal = PreparedValueToSignal.from_origin(
+                    origin=value_to_target
                 )
                 prep_step.prepared_actions.append(prep_value_to_signal)
             except Exception:
@@ -739,13 +743,36 @@ class PreparedValueToSignal:
     name: str
     signal: ophyd.Signal
     value: PrimitiveType
+    origin: ValueToTarget
     result: Result = field(default_factory=incomplete_result)
 
     async def run(self):
+        # generate kwargs for set, exclude timeout and settle time if not provided
+        # in order to use ophyd defaults
+        set_kwargs = {'value': self.value}
+        if self.origin.timeout is not None:
+            set_kwargs.update({'timeout': self.origin.timeout})
+        if self.origin.settle_time is not None:
+            set_kwargs.update({'settle_time': self.origin.settle_time})
+
         try:
-            self.signal.put(value=self.value)
+            status = self.signal.set(**set_kwargs)
+            status.wait()
         except Exception as ex:
             self.result = Result(severity=Severity.error, reason=ex)
             return
 
         self.result = Result()
+
+    @classmethod
+    def from_origin(
+        cls,
+        origin: ValueToTarget
+    ) -> PreparedValueToSignal:
+        pvts = cls(
+            name=origin.name,
+            signal=origin.to_signal(),
+            value=origin.value,
+            origin=origin,
+        )
+        return pvts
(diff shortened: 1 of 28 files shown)


tangkong:
I found a buggy thing I need to fix

ZLLentz:
Some nifty utilities in here
I didn't notice anything odd reading through the code, I'll roll interactively with this for a bit.

atef/widgets/config/utils.py
```diff
@@ -1212,3 +1223,121 @@ def addChild(self, child: TreeItem) -> None:
...
+    add_button: QtWidgets.QToolButton
+    row_label: QtWidgets.QLabel
+
+    def __init__(self, *args, text='Add new row', **kwargs):
+        super().__init__(*args, **kwargs)
+        self.add_button.setIcon(qta.icon('ri.add-circle-line'))
+        self.row_label.setText(text)
+
+
+class TableWidgetWithAddRow(QtWidgets.QTableWidget):
```
ZLLentz, line 1244: Can we disable selection (set the mode to no selection) on these table widgets? It's not very useful here and it makes the rows harder to read by turning them to a darker blut:

image

https://doc.qt.io/qt-6/qabstractitemview.html#selectionMode-prop
https://doc.qt.io/qt-6/qabstractitemview.html#SelectionMode-enum

atef/widgets/config/run_active.py
```diff
@@ -98,3 +101,80 @@ def __init__(self, *args, data: PreparedDescriptionStep, **kwargs):
     def _setup_ui(self) -> None:
         self.title_label.setText(self.bridge.origin.get().name)
         self.desc_label.setText(self.bridge.origin.get().description)
+
+
+class SetValueRunWidget(DesignerDisplay, DataWidget):
```
ZLLentz, line 106: I like all the elements here especially the intuitive options menu, but this widget renders confusingly at large sizes

image

image

I don't have any immediate suggestions about the layout, but I think the reason the large one feels bad is because the two most important bits of information (target, value) end up far apart instead of close together.

If you've had enough of fighting with layouts I can spend some time trying to muck with this too, it's always a maze of size policies...

atef/widgets/config/run_active.py
```diff
@@ -98,3 +101,80 @@ def __init__(self, *args, data: PreparedDescriptionStep, **kwargs):
...
+        for table, field, cls in [
+            (self.actions_table, self.data.prepared_actions, ActionRowRunWidget),
+            (self.checks_table, self.data.prepared_criteria, CheckRowRunWidget)
+        ]:
+            for item in field:
+                ins_ind = table.rowCount()
+                action_row = cls(data=item)
+                table.insertRow(ins_ind)
+                table.setRowHeight(ins_ind, action_row.sizeHint().height())
+                table.setCellWidget(ins_ind, 0, action_row)
```
ZLLentz, line 130: I think the buttons here are good candidates for simple tooltips e.g. "Click here to confirm value" on the value confirmation button

atef/widgets/config/run_active.py
```diff
@@ -98,3 +101,80 @@ def __init__(self, *args, data: PreparedDescriptionStep, **kwargs):
...
+    """
+    filename = 'set_value_run_widget.ui'
+
+    actions_table: QtWidgets.QTableWidget
+    checks_table: QtWidgets.QTableWidget
+
+    def __init__(self, *args, data: PreparedSetValueStep, **kwargs):
+        super().__init__(*args, data=data, **kwargs)
+        self._setup_ui()
+
```
ZLLentz, line 119: One more small thought on a random line: any consideration for whether it'd be more or less intuitive with the value on the left and the arrow pointing to the right? I think that mucks with the layout alignment a bit but it might be worth it

ZLLentz: Alternatively, keeping the ordering and replacing the arrow with text e.g. "set to"? Just so the user can read the line from left to right and make quick sense of it.

All of these random suggestions are just suggestions and I'm not particularly attached to them, but I did feel a little backward looking at this widget and trying to use it for the first time.

tangkong: I was on the fence regarding this for a bit.  I arranged it so the fields would be filled out from left to right; you have to specify a target before you can pick a value to set it to.  

I think it's pretty easy to do this either way

atef/widgets/config/run_active.py
ZLLentz: Maybe a small actionable idea for this is to enforce a maximum arrow length and put a horizontal spacer immediately left of the x, since we don't actually want all the values to run away from their targets, do we?

tangkong: Turns out the a few widgets needed their width policies set to maximum, alongside the horizontal spacer to the right of the delete button 👍

atef/widgets/config/run_active.py
tangkong: Addressed in [93849d0](https://github.com/pcdshub/atef/pull/159/commits/93849d0eee9f6b3d14c1a77d6be658f0e66cce93)

atef/widgets/config/run_active.py
tangkong: addressed in [93849d0](https://github.com/pcdshub/atef/pull/159/commits/93849d0eee9f6b3d14c1a77d6be658f0e66cce93)

atef/widgets/config/utils.py
tangkong: addressed in [93849d0](https://github.com/pcdshub/atef/pull/159/commits/93849d0eee9f6b3d14c1a77d6be658f0e66cce93)

I had originally hooked the selection signal for updating the check row description updates, but ended up finding a different signal to do this

atef/widgets/config/run_active.py
tangkong: Switched the direction of the arrow in [21dbf46](https://github.com/pcdshub/atef/pull/159/commits/21dbf4695b92dbbeeaee659780f8b0492b3afdac)

atef/tests/test_widgets.py
```diff
@@ -210,13 +215,13 @@ def test_config_window_save_load(qtbot: QtBot, tmp_path: pathlib.Path):
         assert source_lines == dest_lines
 
 
-def test_edit_run_toggle(qtbot: QtBot, test_configs: list[os.PathLike]):
+@pytest.mark.skip(reason='inconsistent segfaults on CI, passes locally, (04/17/23)')
```
ZLLentz, line 218: Something I've been meaning to figure out: is there a way to skip tests _only_ on CI? I think in some cases this would be helpful so as to not need to edit the local test suite to run tests that inexplicably brick on GHA.

Maybe the easiest way would be to check for the `CI` environment variable, or maybe there's something better?
https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

tangkong: Seems doable 👀

atef/ui/action_row_widget.ui
```diff
@@ -0,0 +1,170 @@
...
+     <property name="toolButtonStyle">
+      <enum>Qt::ToolButtonTextOnly</enum>
+     </property>
+     <property name="arrowType">
+      <enum>Qt::DownArrow</enum>
+     </property>
+    </widget>
+   </item>
+   <item>
+    <widget class="QToolButton" name="setting_button">
```
klauer, line 54: I really like these widgets! I noticed a couple things while playing around with them:
1. It's possible to have the 'happi item' selector window go in the background. If you then click 'select a target' again, it won't bring that window to the top. If that window is entirely behind the main atef, then it gives the impression that it's broken
2. If I double-click the component name in the happi item window (or even select it and click the 'Select' button):
image
The bottom pop-up goes away and the component setting is unchanged. I have to click it again to show the bottom window and then press 'Apply'. I think that step should be unnecessary here
3. 'Confirm' value has no text?
image

klauer (commented):
I think this is looking good! Lots of little things to address, but it's going in the right direction 👍

atef/tests/test_widgets.py
klauer: It's worth being cautious here and not getting ahead of ourselves to point the finger at CI.

I think there are some threading-related crashes that are the result of actual flaws

atef/procedure.py
```diff
@@ -96,12 +101,103 @@ def allow_verify(self) -> bool:
...
+        return signal
+
+
+@dataclass
+class ValueToTarget(Target):
+    #: the value to set to the target
+    value: Optional[PrimitiveType] = None
+
+    # ophyd.Signal.set() parameters
+    #: connection timeout
```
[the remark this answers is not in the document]
tangkong: yes, my brain for some reason also thought that qualified as a connection.

atef/procedure.py
```diff
@@ -554,3 +651,222 @@ def from_origin(
...
+        Result
+            the step_result for this step
+        """
+        self = cast(SetValueStep, self)
+        for prep_action in self.prepared_actions:
+            action_result = await prep_action.run()
+            if (self.origin.halt_on_fail and action_result.severity > Severity.success):
+                self.step_result = Result(
+                    severity=Severity.error,
+                    reason=f'action failed ({prep_action.name}), step halted'
```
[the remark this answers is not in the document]
tangkong: My intention was for the report generation to access the individual action results, and have step result be an overall summary of sorts.  

I thought that if we had many actions that failed, appending them all to the step result reason would get a bit too lengthy.

atef/procedure.py
```diff
@@ -554,3 +651,222 @@ def from_origin(
...
+        ValueError
+            if the target cannot return a valid signal
+        """
+        signal = origin.to_signal()
+        if signal is None:
+            raise ValueError(f'Target specification invalid: {origin}')
+
+        pvts = cls(
+            name=origin.name,
+            signal=origin.to_signal(),
```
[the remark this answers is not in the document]
tangkong: 🤦

atef/ui/action_row_widget.ui
tangkong: 1: Will be addressed!
2: I actually struggled with this for a long time and couldn't come up with a solution.  I can try again, but my recollection is that changing the default behavior of these QToolButtons required invasive surgery
3: in our rhel fusion style these button boxes always come with icons, which is why I removed the text.   Once we unify the styles (#152)  this should hopefully no longer an issue

klauer: Hmm, I'd be happy to help look into (2). One of the goals of those component-selector widgets was really to be reusable - such that all you have to do is hook into signals to get the information you want. If that's not how it is, then it needs to be fixed

tangkong: This wasn't as bad as I remembered it being 🙄

atef/procedure.py
```diff
@@ -96,12 +101,103 @@ def allow_verify(self) -> bool:
...
+        -------
+        ophyd.EpicsSignal
+            the signal described by this Target
+        """
+        try:
+            if self.device and self.attr:
+                device = util.get_happi_device_by_name(self.device)
+                signal = getattr(device, self.attr)
+            elif self.pv:
+                signal = ophyd.EpicsSignal(self.pv)
```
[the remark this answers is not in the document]
tangkong: Aah I see.  I think I looked at DataCache and prematurely moved on after deciding I didn't want to cache data values.  This seems like exactly what we need here

atef/procedure.py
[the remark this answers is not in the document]
tangkong: I suppose that I didn't consider the action result any more transient than the step result that combines those action results.  If the prepared ProcedureFile exists, the action results, check results, and step results are all accessible.  Placing the action-result messages into the step-result would simply be duplicating the message in multiple places.

I can't imagine a scenario where we would have access to the overall result message but not the action result.  Is that too optimistic of an assumption?

atef/procedure.py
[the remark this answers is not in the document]
tangkong: [3e1b42e](https://github.com/pcdshub/atef/pull/159/commits/3e1b42e198158a679ac15389bc63e956701be2c1)

tangkong:
Re-requested a review from @klauer, at least for a quick look at my responses to his suggestions before leaves on vacation

atef/tests/test_widgets.py
tangkong: Test has been re-enabled, it passes more consistently now though I don't know exactly what helped it get there

atef/procedure.py
```diff
@@ -554,3 +651,222 @@ def from_origin(
...
+        # in order to use ophyd defaults
+        set_kwargs = {'value': self.value}
+        if self.origin.timeout is not None:
+            set_kwargs.update({'timeout': self.origin.timeout})
+        if self.origin.settle_time is not None:
+            set_kwargs.update({'settle_time': self.origin.settle_time})
+
+        try:
+            status = self.signal.set(**set_kwargs)
+            status.wait()  # raises not completed successfully
```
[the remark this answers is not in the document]
tangkong: [3e1b42e](https://github.com/pcdshub/atef/pull/159/commits/3e1b42e198158a679ac15389bc63e956701be2c1)

I think I used this right..

atef/procedure.py
```diff
@@ -554,3 +651,222 @@ def from_origin(
...
+
+
+@dataclass
+class PreparedValueToSignal:
+    #: identifying name
+    name: str
+    #: the signal, derived from a Target
+    signal: ophyd.Signal
+    #: value to set to the signal
+    value: PrimitiveType
```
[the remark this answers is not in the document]
tangkong: #160 has been created to capture this.  I'd like to tackle this outside of this PR, and devote a bit more thought to it

atef/widgets/config/run_base.py
```diff
@@ -275,7 +340,8 @@ def setup_buttons(self, configs, next_widget: AtefItem = None) -> None:
 
     def _make_run_slot(self, configs) -> None:
 
-        def run_slot():
+        @busy_cursor
```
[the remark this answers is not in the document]
klauer: `BusyCursorThread` looks neat! You might want a `try/finally` block around it (along with an `except Exception: raised_exception.emit(ex)` so you can catch it if failed by way of a signal)

I think the desire for wanting blocking behavior is understandable. It just needs to be implemented by other means (that is to say, blocking the GUI thread busily doing non-GUI tasks is truly never the answer).  It may be sensible to disable the specific file pane when such things are running, but you wouldn't want it to interrupt access to other files you have open.

With this running in a background thread now, you could also have a (pretty crucial) hook for stopping checkout tasks that are taking too long for any reason. That'll require some more care/attention to detail, but backgrounding it is I think the first pass at it.

It's a good point about the lack of put completion for PV-only signals. It may be worth adding in the concept of a setpoint PV and a readback PV... For better or worse, that's an essential part of EPICS V3.

atef/procedure.py
```diff
@@ -96,12 +101,103 @@ def allow_verify(self) -> bool:
...
+    """
+    A destination for a value.  Either an ophyd device+attr pair or EPICS PV
+    """
+    #: name of target
+    name: Optional[str] = None
+    #: device name and attr
+    device: Optional[str] = None
+    attr: Optional[str] = None
+    #: EPICS PV
+    pv: Optional[str] = None
```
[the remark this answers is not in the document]
tangkong: #161 created to capture this

atef/procedure.py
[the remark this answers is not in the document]
klauer: ophyd v1 is weird and confusing, it has: 
* `class EpicsSignalBase`, 
* `class EpicsSignal(EpicsSignalBase)`, 
* `class EpicsSignalRO(EpicsSignalBase)`

So while technically correct for your implementation, the generic annotation broke the read-only one used for passive checkouts. 

This is probably fine for now, we can address the typing later

atef/procedure.py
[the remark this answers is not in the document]
klauer: And - apparently "write PV" and "read PV" combinations

klauer:
I think this is looking better as far as the things I pointed out! 

I still see a signal created without tracking or cleanup - need to decide what to do with it:
* https://github.com/tangkong/atef/blob/1da4b6a7017ef7acd1493cb8396710d158f49201/atef/widgets/config/data_active.py#L878-L880

The cache usage seems OK, but I might suggest extending `PreparedSignalComparison` to support `from_signal` so it just uses the signal instance directly instead of referring back to the cache based on pvname (does that make sense?)

atef/procedure.py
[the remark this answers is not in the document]
klauer: I think I lost what my point was here; let's call it resolved 😁

tangkong:
A notes on things I've not yet addressed (and may not in this PR)
- Read/Write PV pairs: This may require a bit more work on the backend.  If we start specifying both read and write pvs, we'll have to extend the `_SignalCache` to either identify signals by the pair.  (While a read/write pair is typically exclusive, can we rule out a user wanting to use multiple different read PV's for the same write PV?)

ZLLentz:
GUI looks pretty nice and is fairly intuitive. I think we could nitpick more but we're nearing the qt diminishing returns point (much effort for small adjustments). The only detail that tripped me up a bit is that the nice table summary in the "Checks" section doesn't update with the check's info until you click on the table row again, otherwise the summary is stale. There are also some discrepancies in the display of these (my equals check has a lot of info in the table, my greater than check does not) but its not a big deal.

Some small things:
- timeout and settle time don't have units
- putting in a disconnect PV very aggressively clears the input box in a way that might be disruptive

I'll give my fullest effort for a readthrough of the code

ZLLentz:
I again didn't see anything odd or wrong in the diff, then again I didn't see anything prior either

tangkong:
It was a bit involved, but I managed to hook up the description update slot to more specific signals in the checks table.  I think I've also addressed the other gui concerns zach brought up.  I'm going to move on to getting some reports set up for the active checkouts, this can probably sit around until Ken comes back for a final approval

tangkong:
I think I'd like to get this merged and tackle followups in other PR's.  I've got some an active checkout report branch waiting to be rebased, and a few bugs to fix before TMO gives it a spin.  

Hoping to get this polished up before I bring shame to my family 🤞

atef/widgets/config/utils.py
```diff
@@ -880,9 +893,20 @@ def make_field(self, value: Any) -> QtWidgets.QWidget:
...
+            int_edit.setMinimum(-1)
+            int_edit.setSpecialValueText('None')
+            int_edit.setToolTip('Input -1 to set value to None')
             int_edit.setValue(value)
             return int_edit
+        elif isinstance(value, float):
+            float_edit = QtWidgets.QDoubleSpinBox()
+            float_edit.setMinimum(-1)
+            float_edit.setSpecialValueText('None')
+            float_edit.setToolTip('Input -1 to set value to None')
```
ZLLentz, line 905: I guess this is as good a way as any to encode `None` into a spinbox

ZLLentz (approved):
I peeked at all the "changes since you last viewed" files and it seems reasonable

atef/widgets/config/data_active.py
```diff
@@ -880,13 +881,12 @@ def confirm_signal(self) -> None:
         try:
             sig.wait_for_connection()
```
klauer, line 881: This is another spot where blocking the GUI thread will easily happen (connect -> block GUI thread -> timeout and raise). This should be moved to a background thread.

PR #164, opened 2023-04-25
ENH: allow Qt args on the command line

### LINKED ISSUE #154: GUI: set 'fusion' style to avoid poor layout on macOS
## Current Behavior

On macOS:

`python -m atef config atef/tests/configs/active_test.json`

image

## Possible Solution

`QT_STYLE_OVERRIDE=fusion python -m atef config atef/tests/configs/active_test.json`

image

This can be done programmatically as in typhos:
https://github.com/pcdshub/typhos/blob/6a5d60ff6b5447ecac03e2723af6e677dbfc007c/typhos/utils.py#L273C4-L275

JJL772:
## Description

Allows Qt command line arguments to be passed through on the command line. Prior to this PR specifying `-style` (or anything else `QGuiApplication` can accept) wasn't possible without argparse complaining.

`-style` is handled specially and will validate against currently available styles, and default to `fusion` on OSX (as per #154).

Closes #154

## Motivation and Context

Qt styling is somewhat operating system specific, which can lead to issues like described in #154. Sometimes it's also desirable to use a different platform plugin or stylesheet. For the most part, these command line arguments are an alternative to envvars Qt also understands.

## How Has This Been Tested?

Tested locally, should work on all systems.

`atef config -style windows`:
Screenshot_20230425_134046

### DIFF
--- a/atef/bin/config.py
+++ b/atef/bin/config.py
@@ -54,7 +54,9 @@ def build_arg_parser(argparser=None):
 
     argparser.description = """
     Runs the atef configuration GUI, optionally with an existing configuration.
-    Qt arguments are also supported. For a full list, see the Qt docs: https://doc.qt.io/qt-5/qguiapplication.html#supported-command-line-options
+    Qt arguments are also supported. For a full list, see the Qt docs:
+    https://doc.qt.io/qt-5/qapplication.html#QApplication
+    https://doc.qt.io/qt-5/qguiapplication.html#supported-command-line-options
     """
 
     argparser.add_argument(
--- a/atef/bin/config.py
+++ b/atef/bin/config.py
@@ -3,10 +3,11 @@
 """
 import argparse
 import logging
+import sys
 from typing import List, Optional
 
 from pydm import exception
-from qtpy.QtWidgets import QApplication
+from qtpy.QtWidgets import QApplication, QStyleFactory
 
 from ..type_hints import AnyPath
 from ..widgets.config.window import Window
@@ -17,6 +18,39 @@ logger = logging.getLogger(__name__)
 def build_arg_parser(argparser=None):
     if argparser is None:
         argparser = argparse.ArgumentParser()
+
+    # Arguments that need to be passed through to Qt
+    qt_args = {
+        '-qmljsdebugger': 1,
+        '-reverse': '?',
+        '-stylesheet': 1,
+        '-widgetcount': '?',
+        '-platform': 1,
+        '-platformpluginpath': 1,
+        '-platformtheme': 1,
+        '-plugin': 1,
+        '-qwindowgeometry': 1,
+        '-qwindowicon': 1,
+        '-qwindowtitle': 1,
+        '-session': 1,
+        '-display': 1,
+        '-geometry': 1
+    }
+
+    for name in qt_args:
+        argparser.add_argument(
+            name,
+            type=str,
+            nargs=qt_args[name]
+        )
+
+    argparser.add_argument(
+        '-style',
+        type=str,
+        choices=QStyleFactory.keys(),
+        default='fusion' if sys.platform == 'darwin' else None,
+        help='Qt style to use for the application'
+    )
 
     argparser.add_argument(
         "filenames",
@@ -29,8 +63,8 @@ def build_arg_parser(argparser=None):
     return argparser
 
 
-def main(filenames: Optional[List[AnyPath]] = None):
-    app = QApplication([])
+def main(filenames: Optional[List[AnyPath]] = None, **kwargs):
+    app = QApplication(sys.argv)
     main_window = Window(show_welcome=not filenames)
     main_window.show()
     exception.install()
--- a/atef/bin/config.py
+++ b/atef/bin/config.py
@@ -21,20 +21,20 @@ def build_arg_parser(argparser=None):
 
     # Arguments that need to be passed through to Qt
     qt_args = {
-        '-qmljsdebugger': 1,
-        '-reverse': '?',
-        '-stylesheet': 1,
-        '-widgetcount': '?',
-        '-platform': 1,
-        '-platformpluginpath': 1,
-        '-platformtheme': 1,
-        '-plugin': 1,
-        '-qwindowgeometry': 1,
-        '-qwindowicon': 1,
-        '-qwindowtitle': 1,
-        '-session': 1,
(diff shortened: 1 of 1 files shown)


tangkong (approved):
I like this.  Quite a nice way of closing #154.  We should probably proliferate this to our other qt apps, but don't take that as me asking you to do it @JJL772 😄

atef/bin/config.py
```diff
@@ -18,6 +19,39 @@ def build_arg_parser(argparser=None):
     if argparser is None:
         argparser = argparse.ArgumentParser()
 
+    # Arguments that need to be passed through to Qt
+    qt_args = {
+        '-qmljsdebugger': 1,
```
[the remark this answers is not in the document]
JJL772: Didn't realize Qt supported GNU style args too, their docs say `-style` and whatnot, but GNU style definitely seems to work. Will fix!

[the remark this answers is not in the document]
JJL772: Using GNU-style args now

atef/bin/config.py
```diff
@@ -18,6 +19,39 @@ def build_arg_parser(argparser=None):
...
+            name,
+            type=str,
+            nargs=qt_args[name]
+        )
+
+    argparser.add_argument(
+        '--style',
+        type=str,
+        choices=QStyleFactory.keys(),
+        default='fusion' if sys.platform == 'darwin' else None,
```
ZLLentz, line 51: Is there any reason why we specifically wanted this only on osx? I know the ticket only mentions mac but maybe it's simpler to just always set it to fusion, but maybe there's some detail I don't know where this breaks something.

typhos just unilaterally applies this as the default (using a different method)
https://github.com/pcdshub/typhos/blob/c1a964aa29e67d35c1ed3ff7c58158a9325e642a/typhos/utils.py#L362

tangkong: A good point.  I don't think it was a fix we only wanted on osx, I interpreted that ticket as the only seeing the issue in osx.  Making the default fusion is probably fine here

atef/bin/config.py
```diff
@@ -29,8 +63,8 @@ def build_arg_parser(argparser=None):
     return argparser
 
 
-def main(filenames: Optional[List[AnyPath]] = None):
-    app = QApplication([])
+def main(filenames: Optional[List[AnyPath]] = None, **kwargs):
+    app = QApplication(sys.argv)
```
ZLLentz, line 67: I'm having a bit of a moment trying to remember how these subparsers work. It looks like **kwargs was added here, then not used, and then the entirety of the cli invocation is passed into QApplication? What exactly is going on? How does QApplication respond if other non-qt-args get into sys.argv?

ZLLentz: Emperically, it looks like kwargs has all the arg values, so it has to be included just to be called, and QApplication seems to run just fine even if you give it extra arguments. Weird stuff.

atef/bin/config.py
```diff
@@ -18,6 +19,39 @@ def build_arg_parser(argparser=None):
...
+            type=str,
+            nargs=qt_args[name]
+        )
+
+    argparser.add_argument(
+        '--style',
+        type=str,
+        choices=QStyleFactory.keys(),
+        default='fusion' if sys.platform == 'darwin' else None,
+        help='Qt style to use for the application'
```
ZLLentz, line 52: Maybe we can include a link to the qt docs as help text either here or somewhere more suitable, to explain all the args when someone calls --help
https://doc.qt.io/qt-5/qguiapplication.html#supported-command-line-options

ZLLentz: I guess we could also just copy paste the text in but that sounds annoying

ZLLentz:
Anyway I think this is a good addition :+1:, implementation-wise it feels like we're just trying to make argparse and qt's own argument parser not fight each other which feels strange but it seems to work.

JJL772:
I'm not an argparse wizard or anything, so I'm open to suggestions on how to do this better

ZLLentz:
I don't think there's an obvious better way to handle this

atef/bin/config.py
JJL772: I ended up linking to the docs in the argparse description string

atef/bin/config.py
JJL772: Sorry, didn't see this yesterday. If we do want to default to fusion, do we want the argument passthrough at all? Initially I was just going to set the style to fusion right after `QApplication` is constructed, but decided on the argument passthrough instead.

tangkong: I think having the option to set the style via cli args is valuable, we can just set the default to fusion regardless of OS

atef/bin/config.py
ZLLentz: Can you add the second link as well?
Oddly enough, the args here are split between those two different docs pages with no linkage between them...

atef/bin/config.py
```diff
@@ -18,6 +19,44 @@ def build_arg_parser(argparser=None):
...
+        type=str,
+        choices=QStyleFactory.keys(),
+        default='fusion',
+        help='Qt style to use for the application'
+    )
+
+    argparser.description = """
+    Runs the atef configuration GUI, optionally with an existing configuration.
+    Qt arguments are also supported. For a full list, see the Qt docs: https://doc.qt.io/qt-5/qguiapplication.html#supported-command-line-options
+    """
```
ZLLentz, line 58:
```suggestion
    Qt arguments are also supported. For a full list, see the Qt docs:
    https://doc.qt.io/qt-5/qapplication.html#QApplication
    https://doc.qt.io/qt-5/qguiapplication.html#supported-command-line-options
    """
```

ZLLentz: Oddly enough, the args here are split between those two different docs pages with no linkage between them...

ZLLentz: I think you're right, which is why it's perplexing that the docs are separated like this...

ZLLentz (approved):
I think this is good to go

ZLLentz:
Required test failure unrelated but I'm going to re-run it anyway

ZLLentz:
That one build is still segfaulting but it must be unrelated to this, merging

atef/bin/config.py
JJL772: I bet their documentation generator doesn't understand the difference between type aliases and the underlying type itself.

PR #166, opened 2023-04-29
FIX: use integers for rect bounds with QIcon.paint

JJL772:
## Description

Typing fix for bounds of `QIcon.paint`. 

## Motivation and Context

This was raising a type error in my conda environment which I setup per the README. I'm not using the pcds conda environment though. I'm thinking this might be related to having Qt6 installed on my system. It seems to absolutely require `int` instead of `float` or another numeric type for `QRect` and `QSize`.

## How Has This Been Tested?

Tested locally in a conda environment with python 3.10.11.

Not documented anywhere

### DIFF
--- a/atef/widgets/config/utils.py
+++ b/atef/widgets/config/utils.py
@@ -8,7 +8,8 @@ from typing import (Any, Callable, ClassVar, Dict, List, Optional, Tuple, Type,
 
 import qtawesome as qta
 from qtpy import QtCore, QtWidgets
-from qtpy.QtCore import QPoint, QPointF, QRectF, QRegularExpression, Qt
+from qtpy.QtCore import (QPoint, QPointF, QRect, QRectF, QRegularExpression,
+                         QSize, Qt)
 from qtpy.QtCore import Signal as QSignal
 from qtpy.QtGui import (QBrush, QClipboard, QColor, QGuiApplication, QPainter,
                         QPaintEvent, QPen, QRegularExpressionValidator)
@@ -535,9 +536,13 @@ class Toggle(QCheckBox):
         # the handle will move along this line
         trailLength = contRect.width() - 2 * handleRadius
         xPos = contRect.x() + handleRadius + trailLength * self._handle_position
-        iconRad = 0.7 * handleRadius
+        iconRad = int(0.7 * handleRadius)
         # center of handle
-        icon_x = xPos - (1.3 * handleRadius) + iconRad
+        icon_x = int(xPos - (1.3 * handleRadius) + iconRad)
+        iconRect = QRect(
+            QPoint(icon_x, round(barRect.center().y()) - iconRad),
+            QSize(2 * iconRad, 2 * iconRad)
+        )
 
         if self.isChecked():
             p.setBrush(self._bar_checked_brush)
@@ -549,8 +554,7 @@ class Toggle(QCheckBox):
             )
             icon = qta.icon(self.checked_icon,
                             color=QColor(self.checked_color).darker())
-            icon.paint(p, icon_x, round(barRect.center().y()) - iconRad,
-                       2 * iconRad, 2 * iconRad)
+            icon.paint(p, iconRect)
 
         else:
             p.setBrush(self._bar_brush)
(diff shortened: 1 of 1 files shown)


atef/widgets/config/utils.py
```diff
@@ -535,9 +536,13 @@ def paintEvent(self, e: QPaintEvent):
         # the handle will move along this line
         trailLength = contRect.width() - 2 * handleRadius
         xPos = contRect.x() + handleRadius + trailLength * self._handle_position
-        iconRad = 0.7 * handleRadius
+        iconRad = int(0.7 * handleRadius)
         # center of handle
-        icon_x = xPos - (1.3 * handleRadius) + iconRad
+        icon_x = int(xPos - (1.3 * handleRadius) + iconRad)
```
tangkong, line 541:
```suggestion
        icon_x = int(xPos - (1.3 * handleRadius) + (1.3 * iconRad))
```

For me this helps center the icon inside the handle, do you mind checking to make sure that holds true in your env?  I don't know why it shouldn't but maybe something weird happens in qt6...

tangkong (commented):
This is fascinating to me.  (This is how you know qt has truly corrupted my mind)

From what I can see in the Qt docs, they've wanted to differentiate between int and float behavior.  The int-accepting `QRect` operates on a pixel level, while the float-y `QRectF` is meant for anti-aliased or scaled content.  So strictly speaking, we shouldn't have been able to put floats.  

I think this is actually a python version issue (or rather how qtpy uses python)  In pcds-conda (py 3.9), you can see the following: 
``` python
In [1]: from qtpy.QtCore import QRect

In [2]: QRect(0.5, 0.5, 0.5, 0.5)
<ipython-input-2-64e96857bd38>:1: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using `__int__` is deprecated, and may be removed in a future version of Python.
  QRect(0.5, 0.5, 0.5, 0.5)
Out[2]: PyQt5.QtCore.QRect()
```

So QRect tries to convert its arguments to ints implicitly, but that gets deprecated at some point in py3.10.  The py3.10 changelog mentions `__int__` [twice](https://docs.python.org/3/whatsnew/3.10.html#other-language-changes), but not in contexts that make immediate sense for this case.  The `__int__` method on `float` isn't itself gone in my py3.10, but at this point I think I can put this down and move on before I get stuck in the rabbit hole

This maybe speaks more to the fact we should update our README.  (Did you install with the 
included `requirements.txt`?  We're pinned to PyQt5 there)  We've been on qt5 due to the desire to support both conda and pypi qtdesigner, I believe.  We may be there for a while, but that doesn't mean we can't make forward-compatible changes like this.

JJL772:
Thanks for digging into this more Robert! Interesting that they removed implicit float -> int conversions, but I guess that's standard for modern languages or something. In C++ I don't think that even produces a compiler warning. 

I did indeed install requirements using `requirements.txt`, so it's definitely using Qt5 still (`qtpy mypy-args` also indicates that). I tried to run atef with `QT_API=pyqt6` but there are some import errors coming from qtpy itself it looks like:
```
ImportError: cannot import name 'Q_ENUMS' from 'qtpy.QtCore' (/home/jeremy/anaconda3/envs/atef/lib/python3.10/site-packages/qtpy/QtCore.py)
```

atef/widgets/config/utils.py
JJL772: As per my other comment...apparently I'm not using Qt6 at all! Sorry for the confusion!

tangkong: I would still suggest this change though, just to center the icon in the handle a bit more.  👍

ZLLentz:
Here's the reference to the Python language change that causes this issue starting in Python 3.10:
https://docs.python.org/3/whatsnew/3.10.html#other-language-changes

> Builtin and extension functions that take integer arguments no longer accept [Decimal](https://docs.python.org/3/library/decimal.html#decimal.Decimal)s, [Fraction](https://docs.python.org/3/library/fractions.html#fractions.Fraction)s and other objects that can be converted to integers only with a loss (e.g. that have the [__int__()](https://docs.python.org/3/reference/datamodel.html#object.__int__) method but do not have the [__index__()](https://docs.python.org/3/reference/datamodel.html#object.__index__) method).

Prior to 3.10, these extension functions would happily cast to `int` for you, but now they will not.
Part of moving our local environments to Python 3.10 is tracking down and fixing all of these instances, which is pretty annoying. Good job catching this one here.

atef/widgets/config/utils.py
JJL772: Sorry for the delay! I've applied this change now

tangkong (approved):
whoops I missed this this time around.  Looks good!  Thanks for catching this

PR #168, opened 2023-05-03
BUG: fix type handling, etc

### LINKED ISSUE #78: BUG: type selection in equality widget seems shaky
## Expected Behavior

It should be easy to save our value as the desired type

## Current Behavior

I had some strange issues where my values got saved as "0.0" string type- I can't remember exactly what I was trying to do in these cases. I think I might have been trying to set 0 integers? Needs investigation.

### LINKED ISSUE #163: BUG: properly handle Sequence types in QDataclassBridge
## Expected Behavior
When encountering a dataclass with type hints involving `Sequence`, we should properly dispatch to `QDataclassList`

## Current Behavior
After moving away from the old `normalize_annotation` typing framework to a `get_origin` based approach, Sequences end up being set to a general `QDataclassValueForobject`.  `get_origin` does not recognize sequences, only basic types so far

``` python
In [2]: get_origin?
Signature: get_origin(tp)
Docstring:
Get the unsubscripted version of a type.

This supports generic types, Callable, Tuple, Union, Literal, Final, ClassVar
and Annotated. Return None for unsupported types. Examples::

    get_origin(Literal[42]) is Literal
```

## Possible Solution
Find a way to recognize sequences...?

* Open up a `ValueSet` comparison page
* try to add a value 

## Context
found in poking around

## Your Environment
pcds-5.7.1

### LINKED ISSUE #165: BUG: AnyValue only accepts string values
## Expected Behavior
If we're comparing a signal to a list of values, we should be able to handle a variety of types

## Current Behavior
Currently if you set the value to a float or otherwise, the type is recognized by the table, but switching to run-mode throws a type error.

The error originates from this [setText call](https://github.com/pcdshub/atef/blob/ddcdf9d74d029c365ff2ea56ae10ce65e543828c/atef/widgets/config/data_passive.py#L1308), we can simply cast the value to a string.  

There's a bit more to this however, even after making this change the comparison doesn't always resolve
image

Still poking around here. 

1. Create an AnyValue comparison
2. Add a float or int value
3. Switch to Run mode and run the comparison

## Context
Was messing around with the GUI making a how-to page and ran into this

## Your Environment
pcds-5.7.1

tangkong:
## Description
Fixes a smattering of bugs I found while messing around with the GUI
* cast values read from the config to a string in AnyValue widget, closes #165 
* Pick up sequences in QDataclassBridge, closes #163 
  * As a consequence transitioning from hard-coded hint recognition, this bug was created.
* Sets the comparison widget type based on the loaded datatype.  closes #78 
* Allows device selection via double-click in the HappiSearchWidget tree-view 
  * Previously there was no way to select a device from this view (no right-click context menu, etc)

## Motivation and Context
Bug free please

## How Has This Been Tested?
Interactively

This PR, and the issues they reference

### DIFF
--- a/atef/qt_helpers.py
+++ b/atef/qt_helpers.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import functools
 import logging
 import platform
+from collections.abc import Sequence
 from typing import (Any, Callable, ClassVar, Dict, Generator, List, Optional,
                     Tuple, Type, get_args, get_origin, get_type_hints)
 
@@ -87,14 +88,16 @@ class QDataclassBridge(QObject):
             # Use dataclass value and override to object type
             NestedClass = QDataclassValue
             dtype = object
-        elif origin is list:
+        elif origin in (list, Sequence):
             # Make sure we have list manipulation methods
+            # Sequence resolved as from collections.abc (even if defined from typing)
             NestedClass = QDataclassList
             dtype = args[0]
         else:
             # some complex Union? e.g. Union[str, int, bool, float]
             # Optional hints also need to have a general signal to emit NoneType
             # (technically QSignal(str) works, but is it worth the special case?)
+            logger.debug(f'Unable to parse type hint: {type_hint} - ({origin}, {args})')
             NestedClass = QDataclassValue
             dtype = object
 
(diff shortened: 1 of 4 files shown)


ZLLentz (approved):
I haven't run this build myself but all the diffs look good

PR #169, opened 2023-05-05
GUI: improve Result refresh behavior

tangkong:
## Description
* Uses signals to notify the GUI to refresh its result icons
* PreparedProcedureGroup now gathers available results from its children before returning the result
* PreparedConfiguration and PreparedGroup also now feature a `combined_result` field and `result` property that calculates the result given its children

## Motivation and Context
I wanted to have the result icons update more dynamically, and also not have to hit run on parent steps after running children steps

I was so stumped on this before and coming back to it with fresh eyes ... really helped/

## How Has This Been Tested?
Interactively so far. Tests have mostly involved opening the gui, opening up a group, running its children, then checking back in with the parent group. 

This PR

### DIFF
--- a/atef/config.py
+++ b/atef/config.py
@@ -342,7 +342,7 @@ class PreparedConfiguration:
     #: The comparisons that failed to be prepared.
     prepare_failures: List[PreparedComparisonException] = field(default_factory=list)
     #: The result of all comparisons.
-    result: Result = field(default_factory=incomplete_result)
+    combined_result: Result = field(default_factory=incomplete_result)
 
     @classmethod
     def from_config(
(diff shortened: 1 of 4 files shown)


atef/config.py
```diff
@@ -457,7 +457,25 @@ async def compare(self) -> Result:
             severity = _summarize_result_severity(GroupResultMode.all_, results)
             result = Result(severity=severity)
 
-        self.result = result
+        self.combined_result = result
+        return result
+
+    @property
+    def result(self):
```
klauer, line 464: Annotation + docstring (two `result` properties in the PR)

klauer (approved):
Looks like a sensible fix to me 👍

ZLLentz (approved):
Looks solid to me

PR #171, opened 2023-05-10
MNT: wrap signal connection check in BusyCursorThread

### LINKED ISSUE #170: MNT: Place signal connection into background thread
This is another spot where blocking the GUI thread will easily happen (connect -> block GUI thread -> timeout and raise). This should be moved to a background thread.

_Originally posted by @klauer in https://github.com/pcdshub/atef/pull/159#discussion_r1187712913_

I think this could be inserted into `BusyCursorThread` easily enough...

tangkong:
## Description
- Places a `signal.wait_for_connection` call into a background thread
- Makes the `BusyCursorThread.raised_exception` actually expect to emit an Exception

## Motivation and Context
Closes #170 

## How Has This Been Tested?
Interactively

This PR

## Pre-merge checklist

### DIFF
--- a/atef/widgets/config/data_active.py
+++ b/atef/widgets/config/data_active.py
@@ -42,7 +42,8 @@ from atef.widgets.config.utils import (ConfigTreeModel, MultiInputDialog,
 from atef.widgets.core import DesignerDisplay
 from atef.widgets.happi import HappiDeviceComponentWidget
 from atef.widgets.ophyd import OphydAttributeData
-from atef.widgets.utils import insert_widget, match_line_edit_text_width
+from atef.widgets.utils import (BusyCursorThread, insert_widget,
+                                match_line_edit_text_width)
 
 from ...procedure import (ComparisonToTarget, DescriptionStep, DisplayOptions,
                           PassiveStep, PlanOptions, PlanStep, ProcedureGroup,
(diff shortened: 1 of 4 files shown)


atef/widgets/config/data_active.py
```diff
@@ -877,15 +878,22 @@ def confirm_signal(self) -> None:
...
-                f'Could not connect to PV: {self.pv_edit.text()}. '
-                'Will be unable to read metadata'
-            )
+
+        def timeout_warning(ex: Exception):
+            if isinstance(ex, TimeoutError):
+                QtWidgets.QMessageBox.warning(
+                    self,
+                    'Failed to connect to PV',
+                    f'Could not connect to PV: {self.pv_edit.text()}. '
```
klauer, line 887: I was going to say: the PV name could be modified within that couple second timeout (since we're no longer blocking the GUI thread after all) and so should be taken out as a variable

... but then I noticed the event filter (`ignore_events=True`). I think that method may need some reworking in the future. Personally, at least, a locked up GUI is maybe my least favorite thing to deal with in UX (which the `BusyCursorThread` should resolve), and a confusingly unresponsive GUI with no option to cancel might be my second least favorite thing (which I think this event filter adds)...

Anyway - `</semi-tangential comment that belongs in an issue>`

klauer (approved):

atef/widgets/config/data_active.py
tangkong: I suppose my approach to preventing those kind of race conditions could be considered heavy-handed, though in this case I'm only disabling a select number of mouse-interaction events.  A more flexible approach could be to provide the thread a specific event filter or button-disable-function.  

I wonder how hard it would be to add a status bar at the bottom of the gui with a cancel-computation button, which wouldn't ever be disabled.  While we're placing tasks in background threads, we don't expose any methods for canceling a step or check.  

I see your point though, I suppose I never ran into the issue during testing since timeouts always resolved quickly enough

ZLLentz (approved):

atef/widgets/config/data_active.py
klauer: > though in this case I'm only disabling a select number of mouse-interaction events.

It's effectively all user interaction that's ignored, isn't it? https://github.com/pcdshub/atef/blob/bf102c36414d543112654e59c57a0f6d3cb64531/atef/widgets/utils.py#L139-L150

I think what you have is a good first pass, to be fair. Doing it properly with threads that cooperate with each other is a really tough thing to do well (in my opinion). This is another sort of scenario where having asyncio for all our signal handling is a good thing; it has built-in support for cancellation whereas threads do not.

> I wonder how hard it would be to add a status bar at the bottom of the gui with a cancel-computation button, which wouldn't ever be disabled. While we're placing tasks in background threads, we don't expose any methods for canceling a step or check.

I think designing with that sort of stuff in mind would be a very good idea

> I see your point though, I suppose I never ran into the issue during testing since timeouts always resolved quickly enough

Yeah, when things go to plan, everything is fine. Until one day the background task hits some unexpected issue and then requires the entire program to be restarted because we can't interact with it any longer...

Final tangential thought: GitHub should allow for transferring discussion threads to an issue for lazy people like me.

tangkong: > It's effectively all user interaction that's ignored, isn't it?

Through the magic of qt you could probably continue typing -> hit enter to submit a new PV.  Also if you were somehow focused on the tree view the arrow keys would still let you navigate, etc.  But yes the average user would probably just see the GUI as locked.

> Final tangential thought: GitHub should allow for transferring discussion threads to an issue for lazy people like me.

I'll transfer this into an issue, I think it's good food for thought

PR #178, opened 2023-05-18
TST: Add tests and improve coverage

tangkong:
## Description
Add tests like I said I would.  I mostly want to see if they pass here, I might remake the PR later

Also brings along a few bug fixes

I'd like to... 
- see about testing button clicks?

## Motivation and Context
I've added a ton of stuff and "neglected" to add tests

## How Has This Been Tested?
These are tests! ✨ 

This PR

## Pre-merge checklist

### DIFF
--- a/atef/report.py
+++ b/atef/report.py
@@ -177,7 +177,7 @@ def build_passive_summary_table(story: List[Flowable], prep_file: PreparedFile):
         if isinstance(item, PreparedConfiguration):
             name = item.config.name
         elif isinstance(item, PreparedComparison):
-            name = item.comparison.name + ' - ' + item.identifier
+            name = str(item.comparison.name) + ' - ' + str(item.identifier)
         name = name or type(item).__name__
         table_data.append(
             [
@@ -951,7 +951,7 @@ class ActiveAtefReport(AtefReport):
             if isinstance(item, PreparedProcedureStep):
                 name = item.origin.name
             elif isinstance(item, PreparedComparison):
-                name = item.comparison.name + ' - ' + item.identifier
+                name = str(item.comparison.name) + ' - ' + str(item.identifier)
             name = name or type(item).__name__
             table_data.append(
                 [
--- a/atef/tests/configs/blank_passive.json
+++ b/atef/tests/configs/blank_passive.json
@@ -1,0 +1,188 @@
+{
+  "version": 0,
+  "root": {
+    "name": "root",
+    "description": null,
+    "tags": null,
+    "configs": [
+      {
+        "ConfigurationGroup": {
+          "name": null,
+          "description": null,
+          "tags": null,
+          "configs": [],
+          "values": {},
+          "mode": "all"
+        }
+      },
+      {
+        "DeviceConfiguration": {
+          "name": null,
+          "description": null,
+          "tags": null,
+          "devices": [],
+          "by_attr": {},
+          "shared": [
+            {
+              "NotEquals": {
+                "name": "untitled",
+                "description": null,
+                "invert": false,
+                "reduce_period": null,
+                "reduce_method": "average",
+                "string": null,
+                "severity_on_failure": 2,
+                "if_disconnected": 2,
+                "value": 0.0,
+                "rtol": null,
+                "atol": null
+              }
+            },
+            {
+              "Greater": {
+                "name": "untitled",
+                "description": null,
+                "invert": false,
+                "reduce_period": null,
+                "reduce_method": "average",
+                "string": null,
+                "severity_on_failure": 2,
+                "if_disconnected": 2,
+                "value": 0.0
+              }
+            },
+            {
+              "GreaterOrEqual": {
+                "name": "untitled",
+                "description": null,
+                "invert": false,
+                "reduce_period": null,
+                "reduce_method": "average",
+                "string": null,
+                "severity_on_failure": 2,
+                "if_disconnected": 2,
+                "value": 0.0
+              }
+            },
+            {
+              "Less": {
+                "name": "untitled",
+                "description": null,
+                "invert": false,
+                "reduce_period": null,
+                "reduce_method": "average",
+                "string": null,
+                "severity_on_failure": 2,
+                "if_disconnected": 2,
+                "value": 0.0
+              }
+            },
+            {
+              "LessOrEqual": {
+                "name": "untitled",
+                "description": null,
+                "invert": false,
+                "reduce_period": null,
+                "reduce_method": "average",
+                "string": null,
+                "severity_on_failure": 2,
+                "if_disconnected": 2,
+                "value": 0.0
+              }
+            },
+            {
+              "Range": {
+                "name": "untitled",
+                "description": null,
+                "invert": false,
+                "reduce_period": null,
+                "reduce_method": "average",
+                "string": null,
+                "severity_on_failure": 2,
+                "if_disconnected": 2,
+                "low": 0,
+                "high": 0,
+                "warn_low": null,
+                "warn_high": null,
+                "inclusive": true
+              }
+            },
+            {
+              "AnyValue": {
+                "name": "untitled",
+                "description": null,
+                "invert": false,
+                "reduce_period": null,
+                "reduce_method": "average",
+                "string": null,
+                "severity_on_failure": 2,
+                "if_disconnected": 2,
+                "values": []
+              }
+            },
+            {
+              "ValueSet": {
+                "name": "untitled",
+                "description": null,
+                "invert": false,
+                "reduce_period": null,
+                "reduce_method": "average",
+                "string": null,
+                "severity_on_failure": 2,
+                "if_disconnected": 2,
+                "values": []
+              }
+            },
+            {
+              "Equals": {
+                "name": "untitled",
+                "description": null,
+                "invert": false,
+                "reduce_period": null,
+                "reduce_method": "average",
+                "string": null,
+                "severity_on_failure": 2,
+                "if_disconnected": 2,
+                "value": 0.0,
+                "rtol": null,
+                "atol": null
+              }
+            }
+          ]
+        }
+      },
+      {
+        "PVConfiguration": {
+          "name": null,
+          "description": null,
+          "tags": null,
+          "by_pv": {},
+          "shared": []
+        }
+      },
+      {
+        "ToolConfiguration": {
+          "name": null,
+          "description": null,
+          "tags": null,
+          "tool": {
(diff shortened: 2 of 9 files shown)


tangkong:
I've spent enough time puttering around here and should probably move onto all those features people want.  I was aiming to increase coverage for each file to >50%, which is almost true.  But in any case I think I added enough example tests and fixtures to make futures tests easier to add.  (hopefully)

atef/tests/configs/blank_passive.json
ZLLentz, line 1: Are these two files "include one of each thing, change nothing" configs? Seems like a good thing to test against

atef/tests/conftest.py
```diff
@@ -108,3 +125,150 @@ def non_interactive_qt_application(monkeypatch):
     monkeypatch.setattr(
         pydm.exception, 'raise_to_operator', lambda *_, **__: None
     )
+
+
+@pytest.fixture
+def load_config():
```
ZLLentz, line 131: Is there a structural reason to prefer passing helper functions like this as fixtures instead of as imports?

tangkong: instead of `from conftest import load_config`?  No, I was probably getting confused messing with fixtures, that might be better honestly.

atef/report.py
```diff
@@ -177,7 +177,7 @@ def build_passive_summary_table(story: List[Flowable], prep_file: PreparedFile):
         if isinstance(item, PreparedConfiguration):
             name = item.config.name
         elif isinstance(item, PreparedComparison):
-            name = item.comparison.name + ' - ' + item.identifier
+            name = str(item.comparison.name) + ' - ' + str(item.identifier)
```
klauer, line 180: Nitpick, but it's just so much easier to read:
```suggestion
            name = f"{item.comparison.name} - {item.identifier}"
```

atef/tests/test_ophyd.py
```diff
@@ -0,0 +1,63 @@
...
+    """ Pass if OphydAttributeData/Summary creation is successful """
+    dev = happi_client.search()[0].get()
+    OphydAttributeData.from_device_attribute(dev, 'setpoint')
+    odata = OphydAttributeData.from_device_attribute(dev, 'acceleration')
+
+    OphydAttributeDataSummary.from_attr_data(odata)
+
+
+# TODO: Figure out how to get sim Enum.  SynGauss has one but requires a device
+# as input ... how to do that in happi?
```
klauer, line 21: I think we could just make a simple enum signal for the test suite here and reference it in happi

atef/tests/test_report.py
```diff
@@ -0,0 +1,34 @@
...
+
+    cache = get_signal_cache()
+    cache.clear()
+    assert len(cache) == 0
+    # Give signals time to be destroyed.  Hypothesis below...
+    # If run in isolation, this can cause CA context unset, presumably caused
+    # by running a callback on a destroyed signal.  Perhaps test suite is exiting
+    # before cleanup is completed?...
+    # If run in the full suite, this test takes place in the middle of the pack,
+    # providing enough time for cleanup?
```
klauer, line 33: Yep, this is a thing for sure.

We could maybe add a fixture for pytest teardown that ensures that everything is cleaned up, like we do in the CLI?

atef/widgets/config/data_active.py
```diff
@@ -982,8 +983,11 @@ class ActionRowWidget(TargetRowWidget):
     setting_button: QtWidgets.QToolButton
     curr_val_thread: Optional[BusyCursorThread]
 
+    edit_widget = ClassVar[Optional[QtWidgets.QWidget]]
```
klauer, line 986:
```suggestion
    edit_widget: ClassVar[Optional[QtWidgets.QWidget]] = None
```

Typo - I've done this one frequently too

Should it be a classvar though? You access it as `self.edit_widget` below

atef/widgets/config/run_active.py
```diff
@@ -155,7 +155,11 @@ def __init__(self, *args, data: PreparedValueToSignal, **kwargs):
         insert_widget(self.status_label, self.status_label_placeholder)
         self.name_label.setText(data.name)
         self.target_label.setText(data.signal.name)
-        self.value_label.setText(str(data.value))
+        enum_strs = getattr(data.signal, 'enum_strs')
+        if enum_strs:
+            self.value_label.setText(str(enum_strs[int(data.value)]))
```
klauer, line 160: Though unlikely, it'd be good to guard against enum string list length mistmatches

That is to say:

```
try:
    self.value_label.setText(str(enum_strs[int(data.value)]))
except IndexError:
    self.value_label.setText(str(data.value))
```

atef/widgets/config/utils.py
```diff
@@ -789,9 +789,11 @@ def get_relevant_pvs(
         for device_name in config.devices:
             dev = util.get_happi_device_by_name(device_name)
             for curr_attr in attrs:
-                pv = getattr(getattr(dev, curr_attr), 'pvname', None)
-                if pv:
+                try:
+                    pv = getattr(getattr(dev, curr_attr), 'pvname', None)
                     pv_list.append((pv, device_name + '.' + curr_attr))
```
klauer, line 794: `getattr(..., "pvname", None)` can return None, and `pv_list` will include a `None` as the first item. Is that OK here?

Also, the `AttributeError` is probably only trying to catch `getattr(dev, cur_attr)`, so the try block should only include that code

klauer (approved):
LGTM, just some additional feedback in comments

atef/tests/test_page.py
```diff
@@ -0,0 +1,161 @@
...
+    comp_page = group_page.full_tree.currentItem().widget
+    old_comp = comp_page.data
+
+    assert isinstance(comp_page, ComparisonPage)
+    new_idx = pick_different_combo_option(comp_page.specific_combo)
+
+    monkeypatch.setattr(QtWidgets.QMessageBox, 'question',
+                        lambda *args, **kwargs: QtWidgets.QMessageBox.Yes)
+    comp_page.specific_combo.setCurrentIndex(new_idx)
+    comp_page.specific_combo.activated.emit(new_idx)
```
ZLLentz, line 158: I'm surprised there isn't an easy way to activate comboboxes via qtbot, it looks like it just gives you mouse and key primitives... oh well

ZLLentz (approved):
I don't have much more to add, other than I think these are are good, well-motivated test suite additions

atef/tests/conftest.py
tangkong: Part of the reason I left it as a helper function was that I expected to write tests that only wanted one config or another, not necessarily all of them.  

I can at least import it as a helper and not a fixture

atef/widgets/config/data_active.py
tangkong: ah good point.  should just be an attribute

atef/widgets/config/utils.py
tangkong: This should not return anything if pvname is none, good catch

atef/tests/test_page.py
tangkong: Also important to note here is that you have to both change the index and emit the activated signal.  I spent too long figuring that out

ZLLentz: right, because `activated` implies that a human clicked on it
https://doc.qt.io/qt-6/qcombobox.html#activated

ZLLentz: this is the sort of thing I'd expect qtbot to be able to do for you 😞

atef/tests/test_report.py
tangkong: I tried a variety of things here, to no immediate success.  Using this fixture and removing the sleep:
``` python
def ophyd_cleanup():
    yield
    dispatcher = ophyd.cl.get_dispatcher()
    if dispatcher is not None:
        dispatcher.stop()
```
* `pytest.fixture(scope='session', autouse=True)`
  * full suite ✅ , test_report.py ❌ (segfault)
* `pytest.fixture(scope='module', autouse=True)`
  * full suite ✅ , test_report.py ❌ (segfault) 
* `pytest.fixture(scope='function', autouse=True)`
  * full suite ❌ (test_at1l0 fails, failing to find the setpoint signal?) , test_report.py ❌ (segfault) 
* `pytest.fixture(scope='function')`, including it only in the report test
  * full suite ✅ , test_report.py ❌ (segfault)

I'm confused and may look back at this later, but I'm going to leave it for now.  Maybe the fixture is wrong?..

klauer: `scope="session"` would have been my first attempt; anything else I'd expect to fail miserably...

Maybe add in a few debug prints to ensure it's happening at the right time?

tangkong: The `scope='session'` inexplicably worked this morning.  🤷

atef/tests/conftest.py
```diff
@@ -118,6 +119,15 @@ def qapp(pytestconfig):
     return application
 
 
+@pytest.fixture(scope='session', autouse=True)
+def ophyd_cleanup():
+    """Clean up ophyd - avoid teardown errors by stopping callbacks."""
+    yield
+    dispatcher = ophyd.cl.get_dispatcher()
```
klauer, line 126: Nitpick - if this ends up working - could we reuse `atef.utils.ophyd_cleanup()` instead?

Also, one more thing - in `atef.tests.test_commandline`, I think we should pass `cleanup=False` to `main()`...

tangkong:
The tests before the most recent failed due to widgets being deleted prematurely.  I suspect that I'm waiting on signals that emit before all the other widgets are finished setting up.  Those other widgets seem to be waiting on CA timeouts that take a while.  (suite fails after test teardown on a callback that runs after we fail to get a signal's value). Perhaps I try reducing that timeout for the test suite, since we don't really expect to connect to whatever PV's we have loaded in the test configs.  ([logs that made me suspicous](https://github.com/pcdshub/atef/actions/runs/5082703114/jobs/9132764618#step:15:747))

The next set of tests pass, but I don't want to leave this in a place where we get intermittent failures depending on race conditions...

tangkong:
Ok I think this is good to go.  Addressed the comments, and slipped in another quick ui bugfix that popped up yesterday.  

If there's no objections I'll merge later today

klauer:
Tests seem to be passing more consistently? I'd say merge at will if you're ready 👍

tangkong:
slightly more consistently.  I suppose we won't know until they fail again 🤷 .

Thanks for the vote of confidence, I'll merge then 👍

PR #183, opened 2023-06-06
BUG: fix observed value report output

tangkong:
## Description
Fixes a quick bug, the observed value could be False-y (e.g. 0), and skip the whole block.  also cleans up try blocks

## Motivation and Context
Make sure reports always report actual observed values 

## How Has This Been Tested?
Interactively 

This PR

## Pre-merge checklist
- [X] Code works interactively

### DIFF
--- a/atef/report.py
+++ b/atef/report.py
@@ -349,15 +349,18 @@ def build_data_table(
     story.append(Paragraph('Observed Data', l0))
     # use cached value.
     observed_value = getattr(comp, 'data', 'N/A')
-    if not observed_value:
+    if observed_value is None:
         # attr can be set to None
         observed_value = 'N/A'
     observed_value = Paragraph(str(observed_value), styles['BodyText'])
     try:
-        timestamp = comp.signal.timestamp.ctime()
-        source = Paragraph(getattr(comp.signal, 'name', ''), styles['BodyText'])
+        timestamp = datetime.fromtimestamp(comp.signal.timestamp).ctime()
     except AttributeError:
         timestamp = 'unknown'
+
+    try:
+        source = Paragraph(getattr(comp.signal, 'name', ''), styles['BodyText'])
+    except AttributeError:
         source = 'undefined'
     observed_data = [['Observed Value', 'Timestamp', 'Source'],
                      [observed_value, timestamp, source]]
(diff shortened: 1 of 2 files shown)


atef/report.py
```diff
@@ -349,15 +349,18 @@ def build_data_table(
...
     try:
-        timestamp = comp.signal.timestamp.ctime()
-        source = Paragraph(getattr(comp.signal, 'name', ''), styles['BodyText'])
+        timestamp = datetime.fromtimestamp(comp.signal.timestamp).ctime()
     except AttributeError:
         timestamp = 'unknown'
+
+    try:
+        source = Paragraph(getattr(comp.signal, 'name', ''), styles['BodyText'])
+    except AttributeError:
```
klauer, line 363: Which part might raise `AttributeError` here?

atef/report.py
```diff
@@ -349,15 +349,18 @@ def build_data_table(
...
     observed_value = getattr(comp, 'data', 'N/A')
-    if not observed_value:
+    if observed_value is None:
         # attr can be set to None
         observed_value = 'N/A'
     observed_value = Paragraph(str(observed_value), styles['BodyText'])
     try:
-        timestamp = comp.signal.timestamp.ctime()
-        source = Paragraph(getattr(comp.signal, 'name', ''), styles['BodyText'])
+        timestamp = datetime.fromtimestamp(comp.signal.timestamp).ctime()
```
ZLLentz, line 357: Just slightly curious: what does running this through datetime first do?

ZLLentz (approved):
Looks good to me!

atef/report.py
```diff
@@ -349,15 +349,18 @@ def build_data_table(
     story.append(Paragraph('Observed Data', l0))
     # use cached value.
     observed_value = getattr(comp, 'data', 'N/A')
-    if not observed_value:
+    if observed_value is None:
```
klauer, line 352: The convenience of omitting `is None` trips me up in these situations, too. I should make more of a point of checking these in reviews

klauer (approved):

atef/report.py
tangkong: Memory fails me, but I think the timestamp may not be a datetime object in some cases, and could just be an integer.  I wanted to use the `.ctime()` method I guess.

atef/report.py
tangkong: If the `PreparedComparison` for some reason doesn't have the `.signal` attribute (missing for PreparedToolComparison` for example)  I should probably touch up those cases...

atef/report.py
tangkong: It's great to be Pythonic, until it isn't... 😢

PR #184, opened 2023-06-14
ENH: Add support for Dynamic Values in Comparisons

tangkong:
## Description
In this episode:
- Adds `DynamicValue`, which queries some external datasource for data to compare a value to.
  - `HappiValue` and `EpicsValue` inherit from `DynamicValue`
- Modifies the `Comparison` classes to have an `async *.prepare()` method, which must be called before `.compare()` can be called
  - This is called when the comparison is prepared (converted to a `PreparedSignalComparison` or otherwise)
- Adds `MultiModeValueEdit`, which can be hooked up for comparisons in active and passive checkouts
- Adds some unit tests
  - Modifies some `test_page.py` tests to try every option in a QComboBox, where applicable.  (The test suite would fail intermittently, and it turned out to be failing whenever we switched to a specific comparison type.)  
  - updates the test configs to include new fields where applicable.  Old configs can still be opened, but the save-load test was failing on the new optional fields
- Report will show prepared `DynamicValue.__str__` as part of the settings table.  This could be pulled out into a separate, more obvious display, but I think this could probably be included in a separate effort.

Notably absent:
- Enabling Dynamic values for `ValueSet`, `AnyValue`
  - These have specific input widgets I would need to mutate, and I want to move on from here

## Motivation and Context
Supersedes #111, but replicates a good bit of it.
I've been getting confused looking at this, so I resorted to referring to values as RHS and LHS.  Prior to this, PR:
- RHS: always static, set and held in `Value` and `Comparison` classes
- LHS: a signal that is read based on the comparison / tool

After this PR:
- RHS: Can be static (as before) or dynamic (pulling data from EPICS or happi at preparation time) 
- LHS: (same as before)

<details> <summary>Various Considerations / Decisions made</summary>

```
* Placing connection logic (signal.get) inside the `Comparison` classes
  * This was done to keep the logic close to the type of `Value`, but if we cared about separating these we would create a 
     set of `Prepared*Value` classes and replace the base `Value` classes at preparation-time.  This was discussed a bit 
     previously, and in the end I avoided it to avoid the refactoring of `PreparedComparison` and its children
* Giving Comparisons a `.prepare()` method
  * This is an accurate name for the method, but I think it might be confusing in the future when someone sees us create a `PreparedComparison` from a `Comparison` that must also call its `.prepare()` method.  If I had more time for navel-gazing I'd think more about renaming things...
```

</details>

## How Has This Been Tested?
Basic unit tests for `EpicsValue` and `HappiValue` have been added

This PR
## Pre-merge checklist
- [x]  Code contains descriptive docstrings, including context and API


atef/check.py
```diff
@@ -83,6 +90,105 @@ def compare(self, value: PrimitiveType) -> bool:
...
+    """
+    # as of 3.10, use kw_only=True to allow mandatory arguments after the inherited
+    # optional ones.  Until then, these must have a default
+    #: The EPICS PV to use.
+    pvname: str = ''
+
+    async def prepare(self, cache: DataCache) -> None:
+        if not self.pvname:
+            raise DynamicValueError('No PV specified')
+        pv_cache = get_signal_cache()
```
klauer, line 155: Shouldn't this just use `cache.get_pv_data()`?

tangkong: Hah you are right d'oh

tangkong:
[6c4a6bd](https://github.com/pcdshub/atef/pull/184/commits/6c4a6bd01a7d27f95bf252d4c8bc6b1d2f949e6c) tests took a couple of retries but passed eventually.  The tests probably create/destroy things too fast and we segfault?  I need to find the right signal/condition to wait for.

tangkong:
I think it's time to move on, in light of all the other feedback/requests rolling in.  The main thing I've skipped is dynamic value support for AnyValue and ValueSet.  I think the `.prepare()` logic might be wrong, but I think there's a lot needed to get it integrated with the GUI currently.  (or maybe I'm just tired of qt)

atef/check.py
```diff
@@ -308,9 +465,39 @@ async def get_data_for_signal_async(
             executor=executor,
         )
 
+    def prepare(self, cache: Optional[DataCache] = None) -> None:
+        """
+        Implement in subclass to grab and cache dynamic values.
+        This is expected to set self.is_prepared to True if
+        successful.
+        """
+        # TODO: think about renaming this method, collides with PreparedComparison
```
klauer, line 474: Beyond naming, it does _feel_ rather awkward when done at this layer.

Without digging into the implementation, I naively would assume that `Comparison` is fairly static (describing the configuration), then `PreparedComparison` wraps the `Comparison` to ready it for dynamic things - connecting to PVs, getting values from the control system - until it can finally execute the comparison and make its judgment. I wouldn't really have expected the signal cache to enter the equation at all in `atef.check`, for example. (I may have even contributed toward messing up the abstraction, so I'm in no way pointing fingers here...)

Regardless, the way it's implemented **works** most importantly!

atef/widgets/utils.py
```diff
@@ -4,10 +4,13 @@
 from typing import ClassVar, Optional
 
 from qtpy import QtCore, QtGui, QtWidgets
-from qtpy.QtCore import QEvent, QObject
-from qtpy.QtGui import QPalette
+from qtpy.QtCore import QEvent, QObject, QRegularExpression
+from qtpy.QtGui import QPalette, QRegularExpressionValidator
 from qtpy.QtWidgets import QLineEdit
 
+PV_regexp = QRegularExpression(r'^\w+(:\w+)+(\.\w+)*$')
```
klauer, line 11: This is partly LCLS-specific (the `:` delimiter, however common it may be even elsewhere)
And also partially incorrect: the field at the end can include JSON or a trailing `$` indicating long string support:
* `PV:NAME.FLD$` (long string)
* `PV:NAME.FLD.{"f":{"lo":0,"hi":10}}` ([channel filter](https://epics.anl.gov/base/R3-15/1-docs/filters.html))
* `PV:NAME.FLD.[1:2]` channel filter array slice shortcut

The TL;DR is that I think regexes for PVs are bound to be wrong, so the best we can/should do is just accept whatever the user types, strip off trailing spaces, and let the control system deal with it.

klauer (commented):
In principle I'm very much 👍 on this PR as I think it's essential functionality for atef.

I want to really try it out (and think about it a bit more) before we move forward, though, so I'll need a bit more time. Let's also let Zach weigh in, of course, while we're away.

atef/check.py
tangkong: It was a point of contention in the previous iteration, for sure.  It is a sort of break in the axioms of atef, or at least the `Normal` - `Prepared` division.

If we were to move the signal-getting into the `PreparedComparison` constructors, those `PreparedComparison`'s would have to know how to handle those unique `DynamicValue`'s, taking that method away from the `Comparison`.  What does the best way of doing that look like?  `isinstance` checks for `HappiValue` and `EpicsValue`?  Another generation of `PreparedComparison` children specific to each `DynamicValue`?   That also felt weird to me, since we're moving the access logic away from the class that holds the access configuration information.  

I'm not super jazzed about it either, but it does work!  It has the interesting side-effect of showing the observed dynamic value in the configuration, as an example of sorts

atef/widgets/utils.py
tangkong: ... aw darn.  I should have known better than to expect that I knew what PVs looked like.  The JSON suffix is  fascinating...

tangkong: Can PVs have spaces at all?  I think the validator can still be helpful in that context

atef/check.py
ZLLentz: > those PreparedComparison's would have to know how to handle those unique DynamicValue's

You're saying that since different comparisons may have different dynamic values applied to various parts of the configuration, the only place this can really be done is here? E.g., otherwise the prepared comparison machinery would have to know which static attributes can be mapped to from each dynamic attribute?

It seems like the only alternative would be to have some other method that lets the prepared comparisons know the mapping from dynamic value to static value to resolve the dynamic behavior themselves, which might be needlessly indirect but would keep all the cache checking in one place.

(as a bonus, such a method could get some mileage in the gui portion of the code too)

atef/check.py
```diff
@@ -308,9 +465,39 @@ async def get_data_for_signal_async(
             executor=executor,
         )
 
+    def prepare(self, cache: Optional[DataCache] = None) -> None:
```
ZLLentz, line 468: Does it matter that this is a function here, but is a coroutine in all the inheriting classes?

atef/widgets/config/utils.py
```diff
@@ -1362,3 +1379,622 @@ def delete_table_row(self, row: QtWidgets.QWidget) -> None:
                 break
 
         self.table_updated.emit()
+
+
+def set_widget_font_size(widget: QWidget, size: int):
```
ZLLentz, line 1384: I'll never stop being surprised that there aren't qt built-ins for stuff like this

atef/widgets/utils.py
ZLLentz: I've never seen a PV with a space in it, but that's likely just to make terminal input less awful

ZLLentz (approved):
I think this all looks clean
I'd run and check this with some prod PVs but everything is down right now
I had a few comments but I don't even remember what they were

atef/widgets/config/utils.py
tangkong: This is one of the things I copied from your old PR, so you were probably just as surprised back then too 😆

atef/check.py
tangkong: You're right, it should be a coroutine.  We never run into problems with it because it's re-defined immediately.

atef/check.py
tangkong: > You're saying that since different comparisons may have different dynamic values applied to various parts of the configuration, the only place this can really be done is here? E.g., otherwise the prepared comparison machinery would have to know which static attributes can be mapped to from each dynamic attribute?

Yes, and we'd end up spawning a new `isinstance` if-else tree, which I'd love to avoid

> It seems like the only alternative would be to have some other method that lets the prepared comparisons know the mapping from dynamic value to static value to resolve the dynamic behavior themselves, which might be needlessly indirect but would keep all the cache checking in one place.

This is a pattern that's been cropping up throughout atef, I think, and it's starting to feel bad from a maintainability point of view.  Every time we add something new, we have to make sure those all-knowing functions get updated as well.  (it's what happened in `atef.widgets.config.page.setup_multi_mode_for_widget`, if I understand your meaning)

Tangentially, the presence of these have made me worry about the overhead associated with adding new steps/comparisons.  If only it were as easy as just defining a new class and making a new widget.  Maybe as I continue to work on atef a cleaner solution will come to me...

atef/widgets/utils.py
tangkong: I'll propose `^\S*$` as the new validator, but will of course wait for Ken to chime in

klauer: The PV part shouldn't have spaces but I'm not sure that applies to the post-`.` JSON blob

ZLLentz:
Some of these older screens may need a rework soon for the potential wide values that come with the EPICS dynamic value
image

ZLLentz:
Some wild desync bug: I was changing the PV list, then changing the selection box for which PV applied, then changing the PV list again, adding and removing rows, etc.
image

ZLLentz:
Note: if you try to run a checkout after one of these syncing bugs, the qt segfaults

ZLLentz:
This is working and feels good use otherwise :+1:

ZLLentz:
One notable omission is something like a "relative" dynamic value, e.g. if I want to check if a motor is within its own limits and use that for 20 motors it's a bit verbose now, but maybe that's what the "templating" thing is supposed to solve?

tangkong:
> Some wild desync bug: I was changing the PV list, then changing the selection box for which PV applied, then changing the PV list again, adding and removing rows, etc. 

I've gotten reports of this in the past but haven't figured out a good way to reproduce and isolate that error.  Do you happen to remember more exactly what the sequence of steps was?

tangkong:
> One notable omission is something like a "relative" dynamic value, e.g. if I want to check if a motor is within its own limits and use that for 20 motors it's a bit verbose now, but maybe that's what the "templating" thing is supposed to solve?

Templating can help, but I saw global values as the solution to this.  We could set up a check with a variable name and set that name at the configuration level, which would fill out the motor name in the rest of the comparisons.  This is pretty hand-wavy, but was how I saw this working.  Maybe this is just how I interpret templating and we're in agreement?

ZLLentz:
Here's the minimal reproduction of the runtime error after some trial and error:
1. Add PV configuration
2. Add a PV to the config
3. Add a comparison
4. Change the PV to a different PV name
5. Select the old PV name from the combo box, causing a key error
6. Delete the comparison
7. Add a comparison

ZLLentz:
> Templating can help, but I saw global values as the solution to this. We could set up a check with a variable name and set that name at the configuration level, which would fill out the motor name in the rest of the comparisons. This is pretty hand-wavy, but was how I saw this working. Maybe this is just how I interpret templating and we're in agreement?

I'm still not sure how I'd go about setting up a dozen motors in one checkout that need to check against motor-specific dynamic values without just making a ton of checks

ZLLentz:
Then again, that doesn't need to be in scope here, or at all

klauer:
> I'm still not sure how I'd go about setting up a dozen motors in one checkout that need to check against motor-specific dynamic values without just making a ton of checks

This was intended to be the framework for dynamic values, right? So hopefully it's solid enough to branch out to "calculated dynamic values" and things along those lines. We had talked about stuff like that along with like "refer to value from previous test" or "refer to value from happi database" or ... well, the list goes on

ZLLentz:
> This was intended to be the framework for dynamic values, right? So hopefully it's solid enough to branch out to "calculated dynamic values" and things along those lines. We had talked about stuff like that along with like "refer to value from previous test" or "refer to value from happi database" or ... well, the list goes on

That's a good point Ken- the framework is solid and extensions are possible if needed.

atef/widgets/utils.py
tangkong: I've removed the validation and added a bunch of `.strip()` calls.  I left the validator "installed" on the PV line edits in the off chance we want to enforce some validation later.  But maybe mostly because of laziness.

tangkong:
I'll tackle the dereferencing error zach found in a separate PR, I've been inspired to write smaller PR's  (see #191)

tangkong:
Any other objections to merging this?  I want to tackle the deallocation error and spacing issues in a separate PR, if possible

ZLLentz:
I think we should merge this for sure

ZLLentz:
In fact, I'm gonna click it

PR #187, opened 2023-06-22
REL: v1.0.0

tangkong:
## Description
I think I did this sem-ver thing right.  Technically since the last tag we changed the config structure, so this was breaking.  We're going with v1.0.0 to simplify our version handling lives.

- generates release notes
- basic docs framework, to be updated later when I have more time

## Motivation and Context
I want to tag this so people can try it out.

## How Has This Been Tested?
These are docs, I ran `make html` locally

These are docs


klauer (approved):

docs/source/releases.rst
```diff
@@ -0,0 +1,59 @@
+Release History
+###############
+
+
+v1.0.0-beta (2023-06-22)
```
ZLLentz, line 5: As far as picking a tag goes, I'd just say we should stick to the rules in https://peps.python.org/pep-0440/ or else the standard packaging tools will explode.

So we shouldn't actually append "beta" to the tag number for python installations... I'm not sure exactly which option in that pep is the right one here.

ZLLentz (commented):
This is really comprehensive.

I think tagging it literally as -beta is clear but it clashes with https://peps.python.org/pep-0440/ and needs at least a little bit of thought.

docs/source/releases.rst
klauer: v1.0.0 is fine, in my opinion

tangkong: We think this is v1.0.0 ready?  I guess I was following a general semver article not the pep one, good looking out.  

I can switch things to v1.0.0

ZLLentz: For the future record, I was apparently wrong about this and a version `v1.0.0-beta` does get normalized to `v1.0.0b0`

tangkong: For what it's worth, it's probably fine to not rely on that normalization if we can avoid it 🤷

PR #193, opened 2023-07-31
ENH: Add a bluesky-plan step

tangkong:
## Description
Adds a step for specifying and running a bluesky step as part of an atef checkout.  Adopts bluesky_queueserver's api for specifying and preparing plans to run, with the eventual dream of being able to support sending plans to a queueserver.  (Whether or not this actually happens, I think their way of handling plans and annotations is worth adopting)

There's a lot that will go into this.  I will enumerate and organize tasks below, and update the list as I decide whether or not to work on any given task.  "completion" is transient at best, as I end up having to re-work previously completed items during development

[Ever-evolving dataclass diagram](https://docs.google.com/drawings/d/1ch44p8pCC4M1C4FldKMPHAMLK_GEFozzgzi6mbdo3Hw/edit?usp=sharing)

I've been told to shift my focus to templated checkouts and other direct stakeholder requests, so I'll be cutting this work short and continuing the effort later (widgets, reports, etc).

### Tasks
<details><summary>Overly detailed task list </summary>
<p>

 - [x] Dataclasses / backend
    - [x] create plan step dataclass
    - [x] create prepared plan step variant
        - [x] needs run uuid, plan "item", prepared comparisons, result
        - [x] fall back to plan names if user does not provide name, needed for plan_id
    - [x] GlobalRunEngine signleton, BlueskyState class
    - [x] check / comparison integration with bluesky plan results
- [x] Programmatic Plan argument widgets, first pass
- [x] Unit Tests
- [x] Supporting work
    - [x] ~get bluesky.plans annotated~  (addressed via atef.annotated_plans)

</p>
</details> 

### thoughts, "tabled" efforts
<details><summary>Details</summary>
<p>

- bluesky-queueserver has a concept of permissions and allowed plans/devices.  For simplicity this effort will assume everyone has access to everything (using a completely permissive `user_groups_permissions.yaml`, but in the future we could filter this with lightpath or specify permissions manually.
- Getting run identifiers sorted out is a bit tricky.  I'll try to describe it below, but it probably needs a visual
    - The requirements are:
        - plans need to be linked to comparisons at EDIT time, meaning they need an identifier that is set and carried over to run/preparation time. -> cannot use RE-generated uuids
        - each checkout file needs access to its own runs, names could be expected to collide between configs/files -> A single global record of plan runs is not possible
        - plan names are not sufficient (plans could be reused), and user-provided names will probably also collide (users are lazy, uniqueness would have to be enforced on GUI side).  -> Need some machinery about generating and saving names
    - As a result, there is a single `BlueskyState` for each `ProcedureFile`, that holds all the run identifiers for that config.
        - plan identifiers are based on the user-provided plan_id, with incrementing suffixes appended
        - these plan identifiers can be accessed via `BlueskyState.run_map` when it comes to letting users choose between existing plans for comparisons
        - all of this is a bit icky, but it works.  

For a separate PR
- [ ] dispatch run engine runs to a non-GUI thread
        - [ ] make pause, stop abort commands available from main thread
        - [ ] add `.clear()` method to reset database with each run-mode preparation?
- [ ] GUI elements
    - [ ] plan building widgets (edit mode)
        - [ ] automated argument entry widgets
    - [ ] plan run widgets (run mode)
    - [ ] plan control (pause, stop, abort) (run)
    - [ ] plan preview utilities (to help reveal what happens in a plan?)
- [ ] Reports
    - [ ] determine how to display run results in report
- [ ] get nabs.plans annotated

</p>
</details> 

## Motivation and Context
Running Bluesky plans is something we want to be available in atef, and the utilities created here will hopefully be useful elsewhere as well.

## How Has This Been Tested?
interactively 

this pr

## Pre-merge checklist

### DIFF
--- a/atef/annotated_plans.py
+++ b/atef/annotated_plans.py
@@ -1,0 +1,574 @@
+"""
+Plans with additional annotations to facilitate gui auto-generation
+In the future this should probably be pulled out into a separate repository
+as it will probably see considerable use
+"""
+from typing import (Any, Callable, Dict, Generator, Iterable, List, Optional,
+                    Union)
+
+import bluesky.plans as bp
+from bluesky_queueserver import parameter_annotation_decorator
+
+
+def add_docstring_from(func):
+    def wrapper(orig_func):
+        if not orig_func.__doc__:
+            orig_func.__doc__ = ""
+        orig_func.__doc__ += (func.__doc__ or "")
+        return orig_func
+
+    return wrapper
+
+
+@parameter_annotation_decorator({
+    "parameters": {
+        "detectors": {"annotation": "typing.List[__DEVICE__]"},
+        "num": {"annotation": "int"},
+        "delay": {"annotation": "typing.Union[typing.Iterable[float], float]"},
+    }
+})
+@add_docstring_from(bp.count)
+def count(
+    detectors: List[Any],
+    num: int = 1,
+    delay: Union[Iterable[Union[float, int]], float, int] = None,
+    *,
+    per_shot: Optional[Callable] = None,
+    md: Optional[Dict[str, Any]] = None
+):
+    yield from bp.count(detectors, num, delay, per_shot=per_shot, md=md)
+
+
+@parameter_annotation_decorator({
+    "parameters": {
+        "detectors": {"annotation": "typing.List[__DEVICE__]"},
+    }
+})
+@add_docstring_from(bp.list_scan)
+def list_scan(
+    detectors: List[Any],
+    *args,
+    per_step: Optional[Callable] = None,
+    md: Optional[Dict[str, Any]] = None
+):
+    yield from bp.list_scan(detectors, *args, per_step=per_step, md=md)
+
+
+@parameter_annotation_decorator({
+    "parameters": {
+        "detectors": {"annotation": "typing.List[__DEVICE__]"},
+    }
+})
+@add_docstring_from(bp.rel_list_scan)
+def rel_list_scan(
+    detectors: List[Any],
+    *args,
+    per_step: Optional[Callable] = None,
+    md: Optional[Dict[str, Any]] = None
+):
+    yield from bp.rel_list_scan(detectors, *args, per_step=per_step, md=md)
+
+
+@parameter_annotation_decorator({
+    "parameters": {
+        "detectors": {"annotation": "typing.List[__DEVICE__]"},
+    }
+})
+@add_docstring_from(bp.list_grid_scan)
+def list_grid_scan(
+    detectors: List[Any],
+    *args,
+    snake_axes: bool = False,
+    per_step: Optional[Callable] = None,
+    md: Optional[Dict[str, Any]] = None
+):
+    yield from bp.list_grid_scan(detectors, *args, snake_axes, per_step=per_step,
+                                 md=md)
+
+
+@parameter_annotation_decorator({
+    "parameters": {
+        "detectors": {"annotation": "typing.List[__DEVICE__]"},
+    }
+})
+@add_docstring_from(bp.rel_list_grid_scan)
+def rel_list_grid_scan(
+    detectors: List[Any],
+    *args,
+    snake_axes: bool = False,
+    per_step: Optional[Callable] = None,
+    md: Optional[Dict[str, Any]] = None
+):
+    yield from bp.rel_list_grid_scan(detectors, *args, snake_axes,
+                                     per_step=per_step, md=md)
+
+
+@parameter_annotation_decorator({
+    "parameters": {
+        "detectors": {"annotation": "typing.List[__DEVICE__]"},
+        "motor": {"annotation": "__DEVICE__"},
+    }
+})
+@add_docstring_from(bp.log_scan)
+def log_scan(
+    detectors: List[Any],
+    motor: Any,
+    start: float,
+    stop: float,
+    num: int,
+    *,
+    per_step: Optional[Callable] = None,
+    md: Optional[Dict[str, Any]] = None
+):
+    yield from bp.log_scan(detectors, motor, start, stop, num,
+                           per_step=per_step, md=md)
+
+
+@parameter_annotation_decorator({
+    "parameters": {
+        "detectors": {"annotation": "typing.List[__DEVICE__]"},
+        "motor": {"annotation": "__DEVICE__"},
+    }
+})
+@add_docstring_from(bp.rel_log_scan)
+def rel_log_scan(
+    detectors: List[Any],
+    motor: Any,
+    start: float,
+    stop: float,
+    num: int,
+    *,
+    per_step: Optional[Callable] = None,
+    md: Optional[Dict[str, Any]] = None
+):
+    yield from bp.rel_log_scan(detectors, motor, start, stop, num,
+                               per_step=per_step, md=md)
+
+
+@parameter_annotation_decorator({
+    "parameters": {
+        "detectors": {"annotation": "typing.List[__DEVICE__]"},
+        "motor": {"annotation": "__DEVICE__"},
+    }
+})
+@add_docstring_from(bp.adaptive_scan)
+def adaptive_scan(
+    detectors: List[Any],
+    target_field: str,
+    motor: Any,
+    start: float,
+    stop: float,
+    min_step: float,
+    max_step: float,
+    target_delta: float,
+    backstep: bool,
+    threshold: float = 0.8,
+    *,
+    md: Optional[Dict[str, Any]] = None
+):
+    yield from bp.adaptive_scan(detectors, target_field, motor, start, stop,
+                                min_step, max_step, target_delta, backstep,
+                                threshold, md=md)
+
+
+@parameter_annotation_decorator({
+    "parameters": {
+        "detectors": {"annotation": "typing.List[__DEVICE__]"},
+        "motor": {"annotation": "__DEVICE__"},
+    }
+})
+@add_docstring_from(bp.rel_adaptive_scan)
+def rel_adaptive_scan(
+    detectors: List[Any],
+    target_field: str,
+    motor: Any,
+    start: float,
+    stop: float,
+    min_step: float,
+    max_step: float,
+    target_delta: float,
+    backstep: bool,
+    threshold: float = 0.8,
+    *,
+    md: Optional[Dict[str, Any]] = None
+):
+    yield from bp.rel_adaptive_scan(detectors, target_field, motor, start, stop,
+                                    min_step, max_step, target_delta, backstep,
+                                    threshold, md=md)
+
+
+@parameter_annotation_decorator({
+    "parameters": {
+        "detectors": {"annotation": "typing.List[__DEVICE__]"},
+        "motor": {"annotation": "__DEVICE__"},
+    }
+})
+@add_docstring_from(bp.tune_centroid)
+def tune_centroid(
+    detectors: List[Any],
+    signal: str,
+    motor: Any,
+    start: float,
+    stop: float,
+    min_step: float,
+    num: int = 10,
+    step_factor: float = 3.0,
+    snake: bool = False,
+    *,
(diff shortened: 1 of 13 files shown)


tangkong:
Test failures at this point are kinda weird.  They seem to consistently fail where they do.

`pydantic.types.conlist` changed its signature with the upgrade to v2, which is the root cause of this error.  This points at the two packages that require pydantic in this environment: databroker and bluesky_queueserver.  databroker<2 passes, while the >2 pre-release versions fail.

tangkong:
As mentioned in the description, I'm leaving it here to address other requests.  There's some jank in here, but hopefully it's mostly sensible.  I'll definitely be revisiting this in the future, as I build the corresponding gui elements

atef/annotated_plans.py
```diff
@@ -0,0 +1,615 @@
...
+@parameter_annotation_decorator({
+    "parameters": {
+        "detectors": {"annotation": "typing.List[__DEVICE__]"},
+        "num": {"annotation": "int"},
+        "delay": {"annotation": "typing.Union[typing.Iterable[float], float]"},
+    }
+})
+@add_docstring_from(bp.count)
+def count(
+    detectors: List[Any],
```
klauer, line 32: Could we do `Any` -> `ophyd.Device` since the decorator overwrites the annotation for the queue server's purposes?

atef/annotated_plans.py
```diff
@@ -0,0 +1,615 @@
...
+    motor_args = [val for pair in zip(motors, point_lists) for val in pair]
+    yield from bp.list_grid_scan(detectors, *motor_args, snake_axes,
+                                 per_step=per_step, md=md)
+
+
+@parameter_annotation_decorator({
+    "parameters": {
+        "detectors": {"annotation": "typing.List[__DEVICE__]"},
+        "motors": {"annotation": "typing.List[__DEVICE__]"},
+        "point_lists": {"annotation": "typing.List[typing.List[float]]"},
```
klauer, line 105: Conversion of `*motor_args` into `list[motors]` and `list[point_lists]` is sensible to me 👍 

However the annotation "override" in the decorator appears to match the one in the type annotation of the plan. Is this line necessary - and if so, why?

atef/enums.py
```diff
@@ -19,3 +19,11 @@ class GroupResultMode(str, enum.Enum):
     all_ = "all"
     #: At least one item must succeed.
     any_ = "any"
+
+
+class PlanDestination(str, enum.Enum):
+    """Destination for plans to be executed"""
+    # local RunEngine
+    local_ = "local"
```
klauer, line 27: Nit: why the `_` suffix on `local_` and `qserver_`?

atef/plan_utils.py
```diff
@@ -0,0 +1,242 @@
...
+        except ValueError:
+            return item, item
+        return device, cpt
+    else:
+        return item, item
+
+
+def get_attr_from_strings(base: Any, cpt: Any) -> Any:
+    """
+    Get ophyd copmonents from the devices in ``base``, if there is an attribute
```
klauer, line 194:
```suggestion
    Get ophyd components from the devices in ``base``, if there is an attribute
```

atef/plan_utils.py
```diff
@@ -0,0 +1,242 @@
...
+        # actually run plan and store uuids
+        logger.debug(f'running plan: {parsed_plan["callable"]}'
+                     f'(*{parsed_args}, **{parsed_kwargs})')
+        run_uuids = self.RE(parsed_plan["callable"](
+                            *parsed_args, **parsed_kwargs))
+        state.run_map[identifier] = run_uuids
+        return run_uuids
+
+
+def separate_attrs_from_strings(item: Any) -> Tuple[Any, Any]:
```
klauer, line 166: (TODO circle back because not sure what these functions do)

atef/plan_utils.py
```diff
@@ -0,0 +1,242 @@
...
+        return {k: get_attr_from_strings(base[k], cpt[k]) for k in base.keys()}
+    elif isinstance(base, list):
+        return [get_attr_from_strings(b, c) for b, c in zip_longest(base, cpt)]
+    elif isinstance(base, ophyd.Device) and (base.name != cpt):
+        return operator.attrgetter(cpt)(base)
+    else:
+        return base
+
+
+def register_run_identifier(state: BlueskyState, name: str) -> str:
```
klauer, line 210: This works on `BlueskyState`, so maybe it should be a method there

The 3rd attempt for registering a name seems like it'd be `new_name_1_2_3` - is that intentional?
Might just want to create a hash, throw it on the end of the "user-defined name" and be done with it

atef/procedure.py
```diff
@@ -21,40 +21,41 @@
...
-        return cls.instance
+# BlueskyState tracker.  Keys are the id of the top-level ProcedureFile
+# Includes one default BlueskyState, with key id(None)
+BS_STATE_MAP: Dict[int, BlueskyState] = {}
 
-    def __init__(self):
-        self.db = databroker.Broker.named('temp')
-        self.RE = RunEngine({})
-        self.RE.subscribe(self.db.insert)
+# Max depth for plan steps
```
klauer, line 57: Depth of 2 would be "plan run inside a plan"?

atef/procedure.py
```diff
@@ -216,27 +242,41 @@ class CodeStep(ProcedureStep):
...
+    #: Name to identify this plan
+    name: str
+    #: The plan name.  Bluesky plan or otherwise
     plan: str
     #: Plan arguments dictionary - argument name to value.
-    args: Sequence[Any]
+    args: Optional[Sequence[Any]] = field(default_factory=list)
     #: Plan keyword  arguments dictionary - argument name to value.
-    kwargs: Dict[Any, Any]
+    kwargs: Optional[Dict[Any, Any]] = field(default_factory=dict)
```
klauer, line 252: Opinion: I find that if you specify `Optional` in the dataclass, you may as well initialize it to `None` instead of default-factory-generated empty dict

atef/procedure.py
```diff
@@ -216,27 +242,41 @@ class CodeStep(ProcedureStep):
...
     #: Arguments which should not be configurable.
-    fixed_arguments: Optional[Sequence[str]]
+    fixed_arguments: Optional[Sequence[str]] = None
+
+    def make_plan_item(self: PlanOptions) -> Dict[str, Any]:
```
klauer, line 256: Nit - many similar functions in dataclasses are named `from_X` or `to_Y` in this codebase

atef/procedure.py
```diff
@@ -878,20 +924,335 @@ def from_origin(
...
+            return Result(
+                severity=Severity.error,
+                reason=('One or more actions failed to prepare: '
+                        f'{[plan.name for plan in self.prepared_plan_failures]}')
+            )
+
+        # get namespace (based on destination?
+        # How will we know what's in the queueserver's destination?)
+        # Run the plans
+        nspace = {}
```
klauer, line 1016:
```suggestion
```

atef/procedure.py
```diff
@@ -878,20 +924,335 @@ def from_origin(
...
+
+def create_prepared_comparison(
+    check: Union[ComparisonToTarget, ComparisonToPlanData],
+    parent: Optional[AnyDataclass] = None
+) -> Union[PreparedComparison, PreparedComparisonException]:
+
+    output = ValueError('Cannot prepare the provided comparison, type not '
+                        f'supported: ({check})')
+
+    if isinstance(check, ComparisonToTarget):
```
klauer, line 1101: You previously mentioned the annoying isinstance if-blocks for these sorts of things, so let me mention something that came to mind: singledispatch
https://docs.python.org/3/library/functools.html#functools.singledispatch
It could be used to remove the `if`/`isinstance` block boilerplate here. I had used it (a total of once) in the pvaccess stuff in caproto and thought it was neat.

atef/procedure.py
```diff
@@ -878,20 +924,335 @@ def from_origin(
...
+                identifier=getattr(check, 'plan_id', ''),
+                message='Failed to initialize comparison',
+                comparison=comp,
+                name=comp.name
+            )
+
+    return output
+
+
+def get_bs_state(dclass: Any):
```
klauer, line 1135: Docstring and annotation would be nice here (not entirely obvious from the name)

atef/procedure.py
```diff
@@ -878,20 +924,335 @@ def from_origin(
...
+    # Use index to grab right uuid, data row
+    uuid = run_uuids[plan_data.plan_no]
+    plan_table: pd.DataFrame = gre.db[uuid].table()
+    field_series: pd.Series = plan_table[plan_data.field_names]
+
+    if isinstance(plan_data.data_points, tuple):
+        data_slice = slice(*plan_data.data_points)
+        data_table = field_series[data_slice].to_numpy()
+    elif isinstance(plan_data.data_points, list):
+        data_table = field_series.iloc[plan_data.data_points].to_numpy()
```
klauer, line 1238: `else: raise ...`?

docs/source/upcoming_release_notes/193-enh_plan_step.rst
```diff
@@ -0,0 +1,24 @@
+193 enh_plan_step
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- Adds backend for running Bluesky plans in active chekckouts
```
klauer, line 10:
```suggestion
- Adds backend for running Bluesky plans in active checkouts
```

klauer (approved):
I know you were asked to work on other things, and there's some nitpicks in my review as per usual. So don't feel the need to fix things up this time around (or at all, they're suggestions after all)

I think this looks like a great step forward 👍

atef/annotated_plans.py
tangkong: An upcoming[ bluesksy queueserver PR](https://github.com/bluesky/bluesky-queueserver/pull/288) should let us do better than `Any`, though I think we could use `ophyd.Device` here since the decorate overrides

atef/annotated_plans.py
tangkong: Ah it's not necessary, you're right.  I got caught autopiloting 😆

atef/procedure.py
tangkong: I was envisioning plan steps inside multiple `ProcedureGroup`s.

atef/plan_utils.py
tangkong: erm, it's supposed to be `new_name_1` -> `new_name_2`, whoops.  I wanted these to be human readable since users would have to choose from a dropdown

atef/procedure.py
tangkong: I guess I wanted it to be optional from a user's point of view (you don't have to provide kwargs at initialization), but it's not optional from the dataclass' point of view (`None` can't be passed as kwargs).  

The right thing to do here is just provide a default, and not mark this as optional

docs/source/upcoming_release_notes/193-enh_plan_step.rst
tangkong: I'm always impressed you have the mental bandwidth to check spelling during code reviews, I don't haha

atef/enums.py
tangkong: This was some combination of me mistakenly thinking `local` was a python keyword and mimicking the preceding `GroupResultMode`.  They don't need the `_` you're right

atef/plan_utils.py
tangkong: Uh, I overcooked this a bit.  At some point during my testing I became convinced that the queueserver could not find components of devices, but it definitely can.  Mass deletion incoming.

atef/procedure.py
ZLLentz: I've never seen this single dispatch thing before, it's neat
shout-outs to the cursed match statement as well for isinstance if trees

ZLLentz (approved):
I'm :+1: on what you have here

PR #198, opened 2023-08-21
ENH: find and replace functionality

tangkong:
## Description
- Implements atef-specific find and replace functionality
- adds a basic find-and-replace widget
- finish up / clean up widget button functionality

## Motivation and Context
This stems from a request for templated checkouts, which boils down to having find-and-replace functionality.  A naive find-and-replace on the json could touch file/field names, and is rather difficult to inspect.  

My intention is to make the future "fill template" step / tool wrap this functionality, possibly with some extra visualization or streamlining buttons.  A sample checkout could be run through this tool and and "filled" as is with this tool, I imagine

(In reality I implemented the simple find-and-replace helpers, tried making it more robust, and a week later ended up here.)

Creating a step that fills templates mid-checkout is a possible next step, though we'd have to consider expanding each template step on preparation and inserting it into the tree.  Inspection of the substeps of that filled template would be un-inspectable unless we expand in edit-mode, and in that case we're basically creating a bunch of filled template checkouts separately.

## How Has This Been Tested?
Some unit tests have been added, more need to be...

This PR

## Screenshots 
image

image

image

## Pre-merge checklist

### DIFF
--- a/atef/ui/fill_template_page.ui
+++ b/atef/ui/fill_template_page.ui
@@ -129,12 +129,12 @@
          <number>0</number>
         </property>
         <item>
-         <widget class="QWidget" name="edits_table_placeholder" native="true">
-          <property name="sizePolicy">
-           <sizepolicy hsizetype="Expanding" vsizetype="Expanding">
-            <horstretch>0</horstretch>
-            <verstretch>0</verstretch>
-           </sizepolicy>
+         <widget class="QLabel" name="edits_table_placeholder">
+          <property name="text">
+           <string>Open a file to get started!</string>
+          </property>
+          <property name="alignment">
+           <set>Qt::AlignLeading|Qt::AlignLeft|Qt::AlignTop</set>
           </property>
          </widget>
         </item>
--- a/atef/ui/fill_template_page.ui
+++ b/atef/ui/fill_template_page.ui
@@ -1,0 +1,202 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>fill_template</class>
+ <widget class="QWidget" name="fill_template">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>570</width>
+    <height>481</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Form</string>
+  </property>
+  <layout class="QVBoxLayout" name="verticalLayout">
+   <item>
+    <layout class="QHBoxLayout" name="horizontalLayout_3">
+     <property name="topMargin">
+      <number>5</number>
+     </property>
+     <item>
+      <widget class="QLabel" name="file_name_label">
+       <property name="text">
+        <string>checkout_name</string>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="QLabel" name="type_label">
+       <property name="text">
+        <string>type_name</string>
+       </property>
+      </widget>
+     </item>
+    </layout>
+   </item>
+   <item>
+    <widget class="QLabel" name="devices_label">
+     <property name="text">
+      <string>Devices in template</string>
+     </property>
+    </widget>
+   </item>
+   <item>
+    <widget class="QListWidget" name="device_list">
+     <property name="maximumSize">
+      <size>
+       <width>16777215</width>
+       <height>50</height>
+      </size>
+     </property>
+    </widget>
+   </item>
+   <item>
+    <layout class="QHBoxLayout" name="horizontalLayout">
+     <property name="topMargin">
+      <number>0</number>
+     </property>
+     <item>
+      <widget class="QFrame" name="edits_frame">
+       <property name="sizePolicy">
+        <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
+         <horstretch>0</horstretch>
+         <verstretch>0</verstretch>
+        </sizepolicy>
+       </property>
+       <property name="minimumSize">
+        <size>
+         <width>50</width>
+         <height>0</height>
+        </size>
+       </property>
+       <property name="frameShape">
+        <enum>QFrame::StyledPanel</enum>
+       </property>
+       <property name="frameShadow">
+        <enum>QFrame::Raised</enum>
+       </property>
+       <layout class="QVBoxLayout" name="verticalLayout_2">
+        <property name="leftMargin">
+         <number>0</number>
+        </property>
+        <property name="topMargin">
+         <number>0</number>
+        </property>
+        <property name="rightMargin">
+         <number>0</number>
+        </property>
+        <property name="bottomMargin">
+         <number>0</number>
+        </property>
+        <item>
+         <widget class="QWidget" name="edits_table_placeholder" native="true">
+          <property name="sizePolicy">
+           <sizepolicy hsizetype="Expanding" vsizetype="Expanding">
+            <horstretch>0</horstretch>
+            <verstretch>0</verstretch>
+           </sizepolicy>
+          </property>
+         </widget>
+        </item>
+       </layout>
+      </widget>
+     </item>
+     <item>
+      <widget class="Line" name="line_3">
+       <property name="sizePolicy">
+        <sizepolicy hsizetype="Fixed" vsizetype="Expanding">
+         <horstretch>0</horstretch>
+         <verstretch>0</verstretch>
+        </sizepolicy>
+       </property>
+       <property name="orientation">
+        <enum>Qt::Vertical</enum>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="QFrame" name="details_frame">
+       <property name="sizePolicy">
+        <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
+         <horstretch>0</horstretch>
+         <verstretch>0</verstretch>
+        </sizepolicy>
+       </property>
+       <property name="minimumSize">
+        <size>
+         <width>50</width>
+         <height>0</height>
+        </size>
+       </property>
+       <property name="frameShape">
+        <enum>QFrame::StyledPanel</enum>
+       </property>
+       <property name="frameShadow">
+        <enum>QFrame::Raised</enum>
+       </property>
+       <layout class="QVBoxLayout" name="verticalLayout_3">
+        <property name="leftMargin">
+         <number>0</number>
+        </property>
+        <property name="topMargin">
+         <number>0</number>
+        </property>
+        <property name="rightMargin">
+         <number>0</number>
+        </property>
+        <property name="bottomMargin">
+         <number>0</number>
+        </property>
+        <item>
+         <widget class="QLabel" name="detail_label">
+          <property name="text">
+           <string>Edit Details</string>
+          </property>
+          <property name="alignment">
+           <set>Qt::AlignCenter</set>
+          </property>
+         </widget>
+        </item>
+        <item>
+         <widget class="QListWidget" name="details_list"/>
+        </item>
+       </layout>
+      </widget>
+     </item>
+    </layout>
+   </item>
+   <item>
+    <layout class="QHBoxLayout" name="horizontalLayout_2">
+     <property name="topMargin">
+      <number>0</number>
+     </property>
+     <item>
+      <widget class="QPushButton" name="open_button">
+       <property name="text">
+        <string>Open File</string>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="QPushButton" name="apply_all_button">
+       <property name="text">
+        <string>Apply All</string>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="QPushButton" name="save_button">
+       <property name="text">
+        <string>Save As...</string>
+       </property>
+      </widget>
+     </item>
+    </layout>
+   </item>
+  </layout>
+ </widget>
+ <resources/>
+ <connections/>
+</ui>
--- a/atef/ui/fill_template_page.ui
+++ b/atef/ui/fill_template_page.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>570</width>
+    <width>750</width>
     <height>481</height>
    </rect>
   </property>
@@ -22,14 +22,14 @@
      <item>
       <widget class="QLabel" name="file_name_label">
        <property name="text">
-        <string>checkout_name</string>
+        <string>[file name]</string>
        </property>
       </widget>
      </item>
      <item>
       <widget class="QLabel" name="type_label">
        <property name="text">
-        <string>type_name</string>
+        <string>[file type]</string>
        </property>
       </widget>
      </item>
--- a/atef/ui/fill_template_page.ui
+++ b/atef/ui/fill_template_page.ui
@@ -15,7 +15,7 @@
   </property>
   <layout class="QVBoxLayout" name="verticalLayout">
    <item>
-    <layout class="QHBoxLayout" name="horizontalLayout_3">
+    <layout class="QHBoxLayout" name="label_hlayout">
      <property name="topMargin">
       <number>5</number>
      </property>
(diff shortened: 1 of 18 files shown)


tangkong:
This is basically feature complete as far as I care for.  There was a good amount of gui-ing here so I'm hoping for some adversarial clicking while I write comments / unit tests.

atef/ui/fill_template_page.ui
ZLLentz, line 1: Clickly clicky seemed to work as I used this and it seems very powerful in a way that I think will be well-appreciated by the users. A few thoughts:
- Subbing into a big structure can sometimes show a bit too much info. Is there any way we can truncate this so we only show the bit that changed?
- Is there any simple way we can automatically refresh when I add a new edit? I got confused for a moment at first before realizing that I needed to click the refresh button on the left.

image

ZLLentz: Oh, also: after saving the file, it seems like this whole template window goes blank. I think that will scare some people. I wonder if we should automatically close this and open the file that was just saved?

tangkong: > Subbing into a big structure can sometimes show a bit too much info. Is there any way we can truncate this so we only show the bit that changed?

I gave this my best effort, but I wonder if I can do any better.  Currently it shows the deepest dataclass, which I thought was needed to give the change any context at all.  But I suppose now that I have {object}.{attr} on the left I could just show the {attr} on the right

> Is there any simple way we can automatically refresh when I add a new edit? I got confused for a moment at first before realizing that I needed to click the refresh button on the left.

This is a perpetual issue I have.  I always try to hook more `editingFinished` signals, but find them to either not catch all the cases I'd want or trigger too often.  I'll see what I can do here

ZLLentz (approved):
I'm very positive on this. Pending Ken's review I think this is good enough to merge.

atef/widgets/config/window.py
```diff
@@ -367,13 +374,39 @@ def clear_results(self, *args, **kwargs):
 
             current_tree.update_statuses()
 
+    def find_replace(self, *args, **kwargs):
+        """ find and replace in the current tree.  Open a find-replace widget """
```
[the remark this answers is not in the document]
ZLLentz: this is really nifty

I wonder if anyone will get confused by the interface though- it wasn't obvious to me at first that this wasn't doing a find/replace in the same window, rather it is creating a whole new version that I could choose to edit and save seperately

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,992 @@
...
+        self.dclass_label.setText(f'{dclass_type}.{attr}')
+
+        pre_text = str(get_item_from_path(self.data.path[:-1], item=self.data.target))
+        # html wrapping to get some line wrapping
+        self.pre_label.setText(f'<font>{pre_text}</font>')
+        self.pre_label.setToolTip(f'<font>{pre_text}</font>')
+        preview_file = copy.deepcopy(self.data.target)
+
+        try:
+            _ = data.apply(target=preview_file)
```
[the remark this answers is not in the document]
tangkong: Ah I dropped my train of thought here.  `FindReplaceAction.apply` catches errors now, so I need that value to decide whether or not application was successful

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,992 @@
...
+        self.match_paths = walk_find_match(self.orig_file, self._match_fn)
+        # generator can be unstable if dataclass changes during walk
+        # this is only ok because we consume generator entirely
+        for path in self.match_paths:
+            action = FindReplaceAction(target=self.orig_file, path=path,
+                                       replace_fn=self._replace_fn)
+
+            self.actions.append(action)
+
+        # this works, but is smelly and bad.  Access parent table signals
```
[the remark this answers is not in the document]
tangkong: The caveat here is that the `TableWidgetWithAddRow` takes care of adding rows to the point where it never exposes those RowWidgets.

Adding a signal on this row that the table widget expects would force me to change the other instances of this table-row-thing to match, which I felt wasn't worth it at this juncture.  I totally agree though, this isn't quite how things should be done.

atef/ui/fill_template_page.ui
tangkong: I've made the previews show just the bit that changed (the last part of the path), as well as gave the user some notice that their file was saved and that they can open it.  

I'll give a bit more information in the details list if the user hasn't picked an edit as well

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+        )
+
+        if not isinstance(edit_row_widget, TemplateEditRowWidget):
+            # placeholder text if nothing is selected
+            l_item = QtWidgets.QListWidgetItem("select an edit or click an edit's"
+                                               "refresh button to show details.")
+            return
+        elif not edit_row_widget.get_details_rows():
+            l_item = QtWidgets.QListWidgetItem(
+                'provide search and replace text to show details...'
```
klauer, line 814: Hmm, it wasn't clear to me why the "edit details" weren't showing:
image
I think this should probably read "Click on edit in the left list to see details; if nothing shows here there is no match" or something to that effect

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+
+    def setup_ui(self):
+        self.child_button.hide()
+        refresh_button = self.button_box.button(QtWidgets.QDialogButtonBox.Retry)
+        refresh_button.clicked.connect(self.refresh_paths)
+        refresh_button.setText('')
+        refresh_button.setToolTip('refresh edit details')
+        refresh_button.setIcon(qta.icon('ei.refresh'))
+
+        apply_button = self.button_box.button(QtWidgets.QDialogButtonBox.Apply)
```
klauer, line 876: If I click 'apply all' after having changed a device in the template, I'd have expected to see the device/signal list above change:
image

atef/widgets/config/window.py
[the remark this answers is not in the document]
klauer: Yeah, it is really cool but I also felt the same regarding the end user's expectations with the dialog.

"Apply All" seems like it should be replacing it in the file, but it's just doing the changes on an in-memory version until you click 'save as' (... right?)
Maybe the dialog should be something like:
> "Choose input file: "
> "Choose output file: "
> (Make edits in this section)
> Reload / Preview and Verify / Save

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+    def reject_action(self) -> None:
+        self.remove_item.emit()
+
+
+class FillTemplatePage(DesignerDisplay, QtWidgets.QWidget):
+
+    file_name_label: QtWidgets.QLabel
+    type_label: QtWidgets.QLabel
+
+    device_table: QtWidgets.QTableWidget
```
klauer, line 567: I think the biggest feature you could add is the ability to change (add/remove/edit) devices and signals directly in this table using the device/component selector dialog:
image

The find/replace stuff is very valuable, but I think the primary thing people will want to template is "test that device instead of this device"

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+        # replace value
+        old_value = parent_item[final_step[1]]
+        parent_item[final_step[1]] = replace_fn(old_value)
+    elif final_step[0] == "__enum__":
+        parent_item = get_item_from_path(path[:-2], item=item)
+        old_enum: Enum = getattr(parent_item, path[-2][1])
+        new_enum = getattr(final_step[1], replace_fn(old_enum.name))
+        setattr(parent_item, path[-2][1], new_enum)
+    else:
+        # simple field paths don't have a final (__sth__, ?) segement
```
klauer, line 225:
```suggestion
        # simple field paths don't have a final (__sth__, ?) segment
```

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+            if isinstance(l_item, get_args(PrimitiveType)) and match(l_item):
+                yield parent + [('__list__', idx)]
+            else:
+                yield from walk_find_match(l_item, match,
+                                           parent=parent + [('__list__', idx)])
+    elif isinstance(item, dict):
+        for d_key, d_value in item.items():
+            # don't halt at first key match, values could also have matches
+            if isinstance(d_value, get_args(PrimitiveType)) and match(d_value):
+                yield parent + [('__dictvalue__', d_key)]
```
klauer, line 100: Checking the coverage - running test suite locally, it doesn't touch this line
(`coverage html && open htmlcov/index.html`)

GUI stuff aside, in my opinion we should try to aim for ~100% coverage on core library stuff that deals with our dataclasses. (Easier said than done of course!)

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+            # cast to float first
+            return int(float(value))
+        else:  # try to cast as original type
+            return type(value)(replace_text)
+
+    return replace_fn
+
+
+def verify_changes(parent, orig_file: Union[ConfigurationFile, ProcedureFile]) -> None:
+    # check to make sure changes are valid
```
klauer, line 251:
```suggestion
    """Check to make sure changes are valid."""
```

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+class FindReplaceAction:
+    target: Union[ConfigurationFile, ProcedureFile]
+    path: List[Tuple[Any, Any]]
+    replace_fn: Callable
+
+    def apply(
+        self,
+        target: Optional[Union[ConfigurationFile, ProcedureFile]] = None,
+        path: Optional[List[Tuple[Any, Any]]] = None,
+        replace_fn: Optional[Callable] = None
```
klauer, line 291:
```python
ReplaceFunction = Callable[...something here...]
```
and
```suggestion
        replace_fn: Optional[ReplaceFunction] = None
```

Would recommend typing the "replace_fn" so you can tell if it's valid

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+        Apply the find-replace action, return True if action was applied
+        successfully.
+
+        Can specify any of ``target``, ``path``, or ``replace_fn`` in order
+        to use that object instead of the stored object
+
+        Parameters
+        ----------
+        target : Optional[Union[ConfigurationFile, ProcedureFile]], optional
+            the file to apply the find-replace action to, by default None
```
klauer, line 303:
```suggestion
            The file to apply the find-replace action to, by default this applies to the current target of the action
```

Or something like that?

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+            logger.debug('failed to open as passive checkout')
+            try:
+                data = deserialize(ProcedureFile, self._original_json)
+            except ValidationError:
+                logger.error('failed to open file as either active '
+                             'or passive checkout')
+
+        return data
+
+    def update_replace_fn(self, *args, **kwargs) -> None:
```
klauer, line 418: Abbreviated method name with star args/kwargs and no docstring... Sorry to be picky (it's genetic, so you're supposed to excuse it I think), but would you mind either expanding out the name or adding brief intent docstrings to these?

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+        self.device_table.setRowCount(max(len(self._signals), len(self._devices)))
+        for i, sig in enumerate(self._signals):
+            self.device_table.setItem(i, 1, QtWidgets.QTableWidgetItem(sig))
+        for i, dev in enumerate(self._devices):
+            self.device_table.setItem(i, 0, QtWidgets.QTableWidgetItem(dev))
+
+    def verify_changes(self) -> None:
+        verify_changes(self, self.orig_file)
+
+    def save_file(self) -> None:
```
klauer, line 713: Shouldn't there be an implicit "verify_changes" call before it's saved?

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+
+    def _fill_devices_list(self) -> None:
+        self.device_table.setRowCount(max(len(self._signals), len(self._devices)))
+        for i, sig in enumerate(self._signals):
+            self.device_table.setItem(i, 1, QtWidgets.QTableWidgetItem(sig))
+        for i, dev in enumerate(self._devices):
+            self.device_table.setItem(i, 0, QtWidgets.QTableWidgetItem(dev))
+
+    def verify_changes(self) -> None:
+        verify_changes(self, self.orig_file)
```
klauer, line 711: Careful, `self.orig_file` can be `None` here per the static analyzer

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+            self._original_json = json.load(fp)
+        try:
+            data = deserialize(ConfigurationFile, self._original_json)
+        except ValidationError:
+            logger.debug('failed to open as passive checkout')
+            try:
+                data = deserialize(ProcedureFile, self._original_json)
+            except ValidationError:
+                logger.error('failed to open file as either active '
+                             'or passive checkout')
```
klauer, line 414: This should probably `raise` and clear `original_json` 
At the very least `data` is unbound in this except clause

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+
+    busy_thread: Optional[BusyCursorThread]
+
+    filename = 'fill_template_page.ui'
+
+    def __init__(
+        self,
+        *args,
+        filepath: Optional[str] = None,
+        window: Optional[Any] = None,
```
klauer, line 590: Probably can do better than `Any` here

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+
+    search_edit: QtWidgets.QLineEdit
+    replace_edit: QtWidgets.QLineEdit
+
+    filename = 'template_edit_row_widget.ui'
+
+    def __init__(
+        self,
+        *args,
+        data=None,
```
klauer, line 855: `data` unused?

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+
+            # clear all rows
+            self.edits_table.clearContents()
+            self.details_list.clear()
+
+    def update_title(self) -> None:
+        """
+        Update the title.  Will be the name and the number of unapplied edits
+        """
+        file_name = os.path.basename(self.fp)
```
klauer, line 774: Dealing with `None` is annoying with respect to the static analyzer but I find it to be good practice - even if you're 100% sure this doesn't get called when `fp is None` - to guard for it anyway. A future refactor you may forget about it and then get confused why `basename(None)` is getting called

atef/widgets/utils.py
```diff
@@ -216,3 +216,93 @@ def reset_cursor(self):
...
+
+    def __init__(self, text: str = "", parent: Optional[QtWidgets.QWidget] = None):
+        super().__init__(parent=parent)
+
+        self._button_text = text
+
+        self.toggle_button = QtWidgets.QToolButton(text=text)
+        self.toggle_button.setCheckable(True)
+        self.toggle_button.setChecked(False)
+        self.toggle_button.setStyleSheet("QToolButton { border: none; }")
```
klauer, line 260: Aside: we should probably move to making an atef stylesheet at some point and move these things there (`atef.qss`, oh yeah!)

klauer (approved):
The functionality here is awesome! Nicely done. I think this will serve the end user rather well.

Some thoughts, suggestions, nitpicks, and musings as per usual. Nothing at a level that requires holding this PR up any longer. Sorry it took me so long to get around to reviewing!

atef/widgets/config/find_replace.py
tangkong: TIL `typing.TYPE_CHECKING` was a thing.  circular imports be gone!

atef/widgets/config/find_replace.py
tangkong: My `TableWidgetWithAddRow` expected its row widgets to be `DataWidget`s, all of which have a `data` kwarg.  I probably could have made a dataclass to represent an "Edit" but apparently past me didn't feel like it 🤷

atef/widgets/utils.py
tangkong: I wonder if that'd fix some of the weird osx rendering differences 🤔 A good idea, but I'm not volunteering right now... 👀

atef/ui/fill_template_page.ui
tangkong: I did think about this, but punted it never to look back.  I think it's honestly the ideal visualization, but involves quite a bit of work I think.  (Particularly because the tree does not often match the dataclass structure)

This is definitely worth a follow up issue methinks

atef/widgets/config/find_replace.py
tangkong: aaaah now that's a fancy idea.  I'll make a followup, because my vision of this quickly became very complicated

atef/widgets/config/find_replace.py
tangkong: I should do this more regularly in my code review as well as devlopment

atef/widgets/config/find_replace.py
tangkong: Don't be sorry, it's entirely your prerogative to to complain about this.  I should be less lazy.

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+            yield from walk_find_match(getattr(item, field.name), match,
+                                       parent=parent + [(item, field.name)])
+    elif isinstance(item, list):
+        for idx, l_item in enumerate(item):
+            # TODO: py3.10 allows isinstance with Unions
+            if isinstance(l_item, get_args(PrimitiveType)) and match(l_item):
+                yield parent + [('__list__', idx)]
+            else:
+                yield from walk_find_match(l_item, match,
+                                           parent=parent + [('__list__', idx)])
```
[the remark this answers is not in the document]
tangkong: This would have probably made things easier.  I'll choose to pass on this, but toss it into an issue if later me decides I have extra cycles

atef/widgets/config/find_replace.py
```diff
@@ -0,0 +1,1009 @@
...
+        if not filename:
+            return
+
+        self.fp = filename
+        self.orig_file = self.load_file(filename)
+        self.setWindowTitle(f'find and replace: ({os.path.basename(filename)})')
+
+    def load_file(self, filepath) -> Union[ConfigurationFile, ProcedureFile]:
+        with open(filepath, 'r') as fp:
+            self._original_json = json.load(fp)
```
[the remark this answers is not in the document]
tangkong: Good catch, this is left over from when I was editing the json directly.  Both here and in the FindReplaceWidget, the original json is never used.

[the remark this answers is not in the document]
tangkong: Also you're totally right we do have loaders. 🤦

atef/widgets/config/find_replace.py
tangkong: I agree with verifying as long as it doesn't prevent saving, since users might use this in an environment where they don't have access to all the necessary devices, making preparation fail.

klauer: Something like "Verification failed! Save anyway? [y/n]"

tangkong:
I think I've addressed all the comments that fit as add-ons to this PR, and made issues for the rest.  I'd like to merge this and move on, unless there are any objections

klauer:
Merge at will!

PR #213, opened 2023-09-29
MNT/BUG: Misc fixes

### LINKED ISSUE #128: BUG: Problematic Relative Tolerance in ATEF Comparison Settings
Was using atef to set the comparison values of a working environmental sensor using the CRIX injection tile and the relative tolerance was set to a value, set to zero without entering, and the relative tolerance indicator would update to 0 despite being equal to the previous set value.

## Expected Behavior
Deleting tolerance value should set the value equal to zero.

## Current Behavior
Setting the relative tolerance value to empty will not update the value to zero.

## Possible Solution
Unsure

## Steps to Reproduce 
Screenshot from 2022-10-04 17-12-00
(for bugs)

## Context
Minor inconvenience

## Your Environment
source pcds_conda

### LINKED ISSUE #195: BUG: Switching the step type may leave behind children tree items
## Expected Behavior
Switching step types should completely change the step type

## Current Behavior
If we start with a SetValueStep, and use the drop-down to change the step, the tree view will still show the child comparison items and clicking on them will go to the ComparisonPage.

Resulting tree layout
image

Switching back to a set value step will remove the child page, and the comparison will not show up in Run-mode, so this bug is low impact. 

## Possible Solution
Go through and regenerate tree after switching?

1. Open the test active config
2. Navigate to the SetValueStep
3. Use the drop-down box to change the step type
4. Existing comparison page will be left behind

## Your Environment
pcds-5.7.3, atef-1.0.0

### LINKED ISSUE #201: Instead of "new file" dialog show "atef welcome"
## Current Behavior
If I open a config through `atef config test.json`, the welcome dialog doesn't show up.
Then if I expect to make a new config, the dialog is less fancy (and perhaps unfamiliar for users)

## Possible Solution
It'd make sense if "New file" just took you to the welcome page, I think

## Context
Nice welcome dialog:
image

Less nice new file dialog:
image

### LINKED ISSUE #202: BUG: atef still attempts to open a file if it cannot be loaded
## Expected Behavior
Atef should just say it can't open the file

## Current Behavior
Attempts to open a new tab with data that doesn't exist

## Possible Solution
Show a message box and return if the deserialization fails

1. Open an old or invalid config

## Context
Was opening old LAS configs

## Your Environment
atef v1.0.0, pcds-5.7.3

### LINKED ISSUE #212: GUI: Range Comparison number line does not update with changed values
## Expected Behavior
Graphical helper line should update as frequently as possible, to be helpful

## Current Behavior
Currently you need to toggle "inclusive" to make this update.

## Possible Solution
This is probably due to the MultiModeValueEdit not emitting the same way past widgets did (or could).  Might look at the `refreshed` signal that's there already

## Context
Was poking around

## Your Environment
pcds-5.7.3, atef-1.2.0

tangkong:
## Description
- `RangeWidget`'s visualizations update a bit more frequently, and also the label text actually updates. Closes #212 
- Adds a menu option to open the welcome tab, since people like it.  Closes #201 
- Properly shows an error message box when a file can't be opened.  Closes #202 
- Removes child `AtefItem` from a ProcedureStep when it's changed from the specific-step-combobox.  Closes #195 
- Allow tolerances to be `None` in Equals comparison.  Modifies the line-edit setup to allow null values ('', None) when casting the line edit value.  Closes #128 

## Motivation and Context
An assortment of fixes because I didn't want to work on anything big but still wanted to be a bit productive

## How Has This Been Tested?
interactively

This PR

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/213-mnt_bug_misc_fixes.rst
+++ b/docs/source/upcoming_release_notes/213-mnt_bug_misc_fixes.rst
@@ -24,4 +24,4 @@ Maintenance
 
 Contributors
 ------------
-- N/A
+- tangkong
--- a/docs/source/upcoming_release_notes/213-mnt_bug_misc_fixes.rst
+++ b/docs/source/upcoming_release_notes/213-mnt_bug_misc_fixes.rst
@@ -1,0 +1,27 @@
+213 mnt_bug_misc_fixes
+######################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- `RangeWidget`'s visualizations update a bit more frequently, and also the label text actually updates. Closes #212
+- Adds a menu option to open the welcome tab, since people like it.  Closes #201
+- Properly shows an error message box when a file can't be opened.  Closes #202
+- Removes child `AtefItem` from a ProcedureStep when it's changed from the specific-step-combobox.  Closes #195
+- Allow tolerances to be `None` in `Equals` comparison.  Modifies the line-edit setup to allow null values (`''`, `None`) when casting the line edit value.  Closes #128
+
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- N/A
--- a/atef/ui/config_window.ui
+++ b/atef/ui/config_window.ui
@@ -37,6 +37,8 @@
     <property name="title">
      <string>File</string>
     </property>
+    <addaction name="action_welcome_tab"/>
+    <addaction name="separator"/>
     <addaction name="action_new_file"/>
     <addaction name="separator"/>
     <addaction name="action_open_file"/>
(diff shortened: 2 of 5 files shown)


docs/source/upcoming_release_notes/213-mnt_bug_misc_fixes.rst
```diff
@@ -0,0 +1,27 @@
...
+- Allow tolerances to be `None` in `Equals` comparison.  Modifies the line-edit setup to allow null values (`''`, `None`) when casting the line edit value.  Closes #128
+
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- N/A
```
ZLLentz, line 27:
```suggestion
- tangkong
```

ZLLentz (approved):
I think all these fixes look correct to me

klauer (dismissed):

ZLLentz (approved):

klauer (approved):

PR #214, opened 2023-10-02
TST: Attempted test suite fixups

tangkong:
## Description
Attempting to fix some intermittent test suite failures.  The crux of the difficulty here is that the suite consistently passes locally, failing only intermittently on GHA CI

## Motivation and Context
- Uses / vendors Typhos' `WeakPartialMethodSlot` for applying callbacks within `ActionRowWidget`
- In one case opted to try making the slots submitted to `BusyCursorThread` actual methods, rather than dynamically created functions
- When pcdsutils is updated, we can remove the vendoring here and rely on pcdsutils' versino

- There is still a very rare test suite failure, but I cannot reproduce it enough to address it.  For the curious, it seems to fail at [`get_curr_value`](https://github.com/pcdshub/atef/actions/runs/6407993392/job/17396405838#step:17:737), the epics-get portion of `ActionRowWidget`
  - While this is the one place we chose to make the callbacks methods rather than partial slots, I think the issue is more likely due to late timeouts than widgets being garbage collected early

## How Has This Been Tested?
GHA, things pass pretty consistently locally 

This PR

## Pre-merge checklist

### DIFF
--- a/atef/widgets/config/data_active.py
+++ b/atef/widgets/config/data_active.py
@@ -25,7 +25,7 @@ import qtawesome
 import typhos
 import typhos.cli
 import typhos.display
-from ophyd.signal import EpicsSignalBase
+from ophyd.signal import Signal
 from qtpy import QtCore, QtGui, QtWidgets
 from qtpy.QtCore import Qt
 from qtpy.QtWidgets import QDialogButtonBox
@@ -901,7 +901,7 @@ class ActionRowWidget(TargetRowWidget):
         self.edit_widget = None
         if data is None:
             data = ValueToTarget()
-        self._sig = None
+        self._sig: Optional[Signal] = None
         self._curr_value = None
         self._dtype = None
         self._enum_strs = None
@@ -960,7 +960,7 @@ class ActionRowWidget(TargetRowWidget):
         Updates value input widget with a QLineEdit with the approriate validator
         given the target's datatype
         """
-        self._sig: EpicsSignalBase = self.data.to_signal()
+        self._sig = self.data.to_signal()
         if self._sig is None:
             self.edit_widget = QtWidgets.QLabel('(no target set)')
             insert_widget(self.edit_widget, self.value_input_placeholder)
--- a/atef/widgets/config/data_active.py
+++ b/atef/widgets/config/data_active.py
@@ -901,6 +901,10 @@ class ActionRowWidget(TargetRowWidget):
         self.edit_widget = None
         if data is None:
             data = ValueToTarget()
+        self._sig = None
+        self._curr_value = None
+        self._dtype = None
+        self._enum_strs = None
         super().__init__(data=data, **kwargs)
 
     def setup_ui(self) -> None:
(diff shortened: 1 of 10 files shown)


tangkong:
Since I mistakenly left an `assert False` statement, gha now sees what I was seeing locally.  While all the tests fail, some of them fail + "attempt to access deleted object"

[py3.11, pip](https://github.com/pcdshub/atef/actions/runs/6398026024/job/17367254264?pr=214#step:14:836) fails when trying to teardown some PageWidgets, though test suite completes

[py3.9, conda](https://github.com/pcdshub/atef/actions/runs/6398026024/job/17367253203?pr=214#step:17:738) fails with a straight up segfault, suite does not complete

tangkong:
A slightly better result, though failing the first time ([py3.10, conda](https://github.com/pcdshub/atef/actions/runs/6398074094/job/17367403320?pr=214#step:17:826))

tangkong:
Replacing all `partial` callbacks with `WeakPartialMethodSlot` cleans up the test suite quite nicely.  

I may try to add the class to pcdsutils before I merge this, for the sake of not having to return to this

tangkong:
I'm not sure if I'm supposed to keep hard references to the `WeakPartialMethodSlot` object.  I got a single segfault testing locally, but haven't been able to get one since, so I'm not sure if it's actually a problem.  The GHA test suite seems to be passing without issue, so that's not really helpful

tangkong:
This is ready to be reviewed and merged, with pcdsutils getting WeakPartialMethodSlot

atef/widgets/config/data_active.py
```diff
@@ -927,13 +931,37 @@ def update_target(self) -> None:
...
+
     def update_input_placeholder(self) -> None:
         """
         Updates value input widget with a QLineEdit with the approriate validator
         given the target's datatype
         """
-        sig: EpicsSignalBase = self.data.to_signal()
-        if sig is None:
+        self._sig: EpicsSignalBase = self.data.to_signal()
+        if self._sig is None:
```
ZLLentz, line 964: This bit is confusing from a type hinting perspective and probably gives some of the stricter type checking libraries pause. It's saying "if this thing that is always an EpicsSignalBase is None..."

Maybe this type should be `EpicsSignalBase | None` and maybe it should be hinted in the object init or in the class definition itself instead of in this method?

ZLLentz (dismissed):
I think all these changes are clean and well-motivated
One nitpick about type annotations

atef/widgets/config/data_active.py
klauer: It might also just be a regular signal, right?

klauer (dismissed):
Same as @ZLLentz - a bit of class body type hints are probably in order here.

atef/widgets/config/data_active.py
tangkong: Turns out `atef.Procedure.Target.to_signal() -> Optional[EpicsSignal]` as written currently, so I'll keep it as such and move it to the init

ZLLentz (approved):

Issue #160, opened 2023-04-19
ENH: Consider non-primitive data types


tangkong:
## Expected Behavior

Atef should be able to handle non-primitive (not int, float, bool, or str) data types for both passive and active checkouts.  
- Passive checkouts should allow checking values in data such as arrays
- Active checkouts should support setting, say, lists to a PV

## Current Behavior

`Target` and its subclasses used in `SetValueStep` expect only `PrimitiveType`.
All `Value` subclasses expect their `value` field to be a [PrimitiveType](https://github.com/pcdshub/atef/blob/073b8f47833279fd0ea93a5547bfd4ef2d7ba1f8/atef/check.py#L42)

This assumption is reflected in our GUI, which only really has simple LineEdits and SpinBoxes for input widgets

## Possible Solution

🤔 
Expanding the existing dataclasses to handle more complex data types will probably require some more involved logic in comparison.  QDataclassBridges will also need to be updated to handle a potential `Union[List[Any], PrimitiveType]`

Obviously our GUI framework will have to extend as well, a natural list input widget will need to developed and applied more programatically

## Context

Brought up in #159 

## Your Environment
pcds-5.7.1

klauer:
For passive checkouts, our initial pass could be summarized as: "you must reduce your array data into a scalar to be able to compare [it and use it in a passive check]"

Is that still a reasonable approach for passive checks? I think it's worth discussing.

For active checks, it's not unreasonable to put an array (e.g., of calibration coefficients or something) and verify that they get set properly.

PR #219, opened 2023-10-18
MNT: make enum checks more robust

### LINKED ISSUE #200: Enum values are not checked fully correctly
## Expected Behavior

If I make a check that runs against an enum value, it should be converted to the enum string before being compared.

## Current Behavior

Enum values are kept as integers for the comparison:

image

## Possible Solution

I guess we need to convert these or store both the int and the enum when we get the values

1. make a check against an enum
2. run the checkout

## Context

I was click testing #198 when I noticed this

## Your Environment

pcds-5.7.3 + atef from #198

### LINKED ISSUE #220: BUG: string / enum comparison fails if tolerances are set
## Expected Behavior
Tolerance should be ignored intelligently

## Current Behavior
If you set tolerances and change the type back to a one where tolerances don't make sense, the tolerances will persist, and an equals check will try to use them

## Possible Solution
Clear tolerance information when switching away from float-mode.

1.  Set up a working enum comparison
2. Change the type to float, and set non-None tolerances
3. Change the type back to enum, and comparison will fail

## Context
Poking around 

## Your Environment
pcds-5.7.3, atef-1.2.0

tangkong:
## Description
When checking against an enum signal, check against both the integer and string value if needed.
Also ignore tolerances unless comparing against a number

## Motivation and Context
closes #200 
closes #220 

## How Has This Been Tested?
Interactively, and a test case has been added. 

This PR

## Pre-merge checklist

### DIFF
--- a/atef/check.py
+++ b/atef/check.py
@@ -87,12 +87,14 @@ class Value:
 
     def compare(self, value: PrimitiveType) -> bool:
         """Compare the provided value with this one, using tolerance settings."""
-        if self.rtol is not None or self.atol is not None:
+        if ((self.rtol is not None or self.atol is not None)
+                and isinstance(value, (int, float))):
             return np.isclose(
                 value, self.value,
                 rtol=(self.rtol or 0.0),
                 atol=(self.atol or 0.0)
             )
+
         return value == self.value
 
 
--- a/atef/check.py
+++ b/atef/check.py
@@ -88,7 +88,7 @@ class Value:
     def compare(self, value: PrimitiveType) -> bool:
         """Compare the provided value with this one, using tolerance settings."""
         if ((self.rtol is not None or self.atol is not None)
-                and isinstance(value, (int, float))):
+                and not isinstance(value, (str, bool))):
             return np.isclose(
                 value, self.value,
                 rtol=(self.rtol or 0.0),
--- a/atef/config.py
+++ b/atef/config.py
@@ -1207,6 +1207,17 @@ class PreparedSignalComparison(PreparedComparison):
             identifier=self.identifier
         )
 
+    async def compare(self) -> Result:
+        # if signal is an enum, also flip the string field and retry
+        result = await super().compare()
+
+        if result.severity != Severity.success and self.signal.enum_strs:
+            self.comparison.string = not self.comparison.string
+            await self.get_data_async()
+            result = await super().compare()
+
+        return result
+
     @classmethod
     def from_device(
         cls,
--- a/atef/config.py
+++ b/atef/config.py
@@ -1214,7 +1214,8 @@ class PreparedSignalComparison(PreparedComparison):
         """
         result = await super().compare()
 
-        if result.severity != Severity.success and self.signal.enum_strs:
+        if (result.severity != Severity.success
+                and getattr(self.signal, 'enum_strs', None)):
             self.comparison.string = not self.comparison.string
             await self.get_data_async()
             result = await super().compare()
--- a/atef/config.py
+++ b/atef/config.py
@@ -1208,7 +1208,10 @@ class PreparedSignalComparison(PreparedComparison):
         )
 
     async def compare(self) -> Result:
-        # if signal is an enum, also flip the string field and retry
+        """
+        Run the inherited compare routine, but if signal is an enum,
+        also flip the string field and retry
+        """
         result = await super().compare()
 
         if result.severity != Severity.success and self.signal.enum_strs:
--- a/atef/config.py
+++ b/atef/config.py
@@ -1207,21 +1207,6 @@ class PreparedSignalComparison(PreparedComparison):
             identifier=self.identifier
         )
 
-    async def compare(self) -> Result:
-        """
-        Run the inherited compare routine, but if signal is an enum,
-        also flip the string field and retry
-        """
-        result = await super().compare()
-
-        if (result.severity != Severity.success
-                and getattr(self.signal, 'enum_strs', None)):
-            self.comparison.string = not self.comparison.string
-            await self.get_data_async()
-            result = await super().compare()
-
-        return result
-
     @classmethod
     def from_device(
         cls,
--- a/atef/reduce.py
+++ b/atef/reduce.py
@@ -23,10 +23,13 @@ class EnumValue:
     str_value: str
 
     def __eq__(self, __value: object) -> bool:
-        if isinstance(__value, int):
-            return self.int_value == __value
-        elif isinstance(__value, str):
-            return self.str_value == __value
+        if isinstance(__value, (int, str)):
+            try:
+                int_val = int(__value)
+            except ValueError:
+                int_val = __value
+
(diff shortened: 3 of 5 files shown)


atef/config.py
```diff
@@ -1207,6 +1207,21 @@ async def _compare(self, data: Any) -> Result:
...
+    async def compare(self) -> Result:
+        """
+        Run the inherited compare routine, but if signal is an enum,
+        also flip the string field and retry
+        """
+        result = await super().compare()
+
+        if (result.severity != Severity.success
+                and getattr(self.signal, 'enum_strs', None)):
+            self.comparison.string = not self.comparison.string
```
klauer, line 1219: This sets it and never resets it? That seems not good

It would seem like to me that we could be a bit smarter and get the string value from the enum strings list ourselves and redo the comparison
"If the comparison failed and we have an integer from the control layer, get the corresponding string value and run a new simple comparison. If and only if that succeeds, adjust the `result` (otherwise stick with the original failure)"

tangkong: We should definitely reset the value, but I wanted to keep the data-gathering logic out of this class to avoid fragmenting that logic across multiple locations.

klauer: You're already digging into the signal to pull out enum strings in the comparison - that seems to be reaching beyond where it should.

This specific PR is trying to figure out the user's intent ("oh they must not know what they're asking for, let's try the other method"). The original idea was to make the user specify whether they wanted a string comparison or an integer comparison by selecting "string". What happens if there's some overlap/ambiguity in the values? e.g., enum strings `["0", "2", "foobar"]`, and the comparison wants to know if the value is "2"? I suppose either "2" or "foobar" would be valid in that case, rather strangely.

If that's still not a concern, perhaps it'd be worthwhile to consider a special wrapper for enum values at the signal caching / value retrieval layer. That is to say a wrapper that says `1 == "Enum Value for 1"` is `True`

atef/config.py
```diff
@@ -1207,6 +1207,21 @@ async def _compare(self, data: Any) -> Result:
...
+            await self.get_data_async()
+            result = await super().compare()
```
ZLLentz, line 1221: Will this give us confusing results in the cases where both versions of the check fail?

atef/config.py
tangkong: To me there was a distinction between checking if enum_strs exists on the signal (which is already stored on the `PreparedSignalComparison`) and using the cache to get the signal value (`self.cache.get_signal_data`).  Are those two less distinct than I thought?  

----------------------------------

The `["0", "2", "foobar"]` case is interesting, and I hope nobody makes an enum like this...  A bit of me detailing my thought process below.
If the user wants to check the signal against `"2"` (string), getting the data from the signal will only return one of the enum values.  
* if signal has "2" (index 1), the signal could either return `"2"` (string) or `1` (int)
  * comparison passes on the first case
* if signal has "foobar" (index 2), the signal could either return `"foobar"` (string) or `2` (int)
  * comparison fails on both cases 

The key here is that we're only ever comparing to an enum in one of the positions.

If we do try to grab the enum value corresponding to the user provided "2", then the potential for confusion arises.  The comparison's "string" setting only changes what we get from the signal, not the user's provided value.  

I think the upshot of the above block of text is I feel like the current process isn't too ambiguous.

----------

Having finally made it to the last bit of your comment I like that idea better than all the above 🤣

ZLLentz (dismissed):
I think this is a pretty solid implementation

tangkong:
I'm going to try to tackle #220  here as well, but have gotten lost in the maze of qt intricacies.  I'll re-request review when I've sorted myself out

klauer (approved):

tangkong:
The test suite is being exceptionally finicky.  As usual it passes locally

tangkong:
All that's changed since the last review is a quick  numeric-check before we call np.isclose.  This ended up being the easiest way to deal with irrelevant tolerances.  (I did not fix the qt issues... yet)

atef/check.py
```diff
@@ -87,12 +87,14 @@ def get(self) -> PrimitiveType:
 
     def compare(self, value: PrimitiveType) -> bool:
         """Compare the provided value with this one, using tolerance settings."""
-        if self.rtol is not None or self.atol is not None:
+        if ((self.rtol is not None or self.atol is not None)
+                and isinstance(value, (int, float))):
```
klauer, line 91: Maybe a `not isinstance(value, (str, bool))` would be better, considering how `int`/`float`-like values tend to be more common.

atef/reduce.py
```diff
@@ -13,6 +13,27 @@
...
+class EnumValue:
+    """
+    Wrapper for enum value that can be compared to either the string or int
+    value of the enum.  Only accepts string enums.
+    """
+    int_value: int
+    str_value: str
+
+    def __eq__(self, __value: object) -> bool:
+        if isinstance(__value, int):
```
klauer, line 26: `return __value == self.int_value or __value == self.str_value`?

Since we're trying to help out the user here with common mistakes, is there a chance `__value` comes in erroneously as like `"3"` instead of (int) `3` and we need to do a `int(__value) == self.int_value`?

klauer (dismissed):
Seems good to me 👍

atef/check.py
tangkong: I guess if we assume only PrimitiveType (str, int, float, bool) will be submitted, this is equivalent.  Though is this really a significant time save?

klauer: > if we assume only PrimitiveType (str, int, float, bool) will be submitted

I may be totally wrong here but I think `PrimitiveType` is more of a _hope_ than perhaps a reality. My recollection of my code here isn't as good as I'd hope, so let me ask: could `np.ndarray` leak in here if we don't have a reduction method set? Could values directly from the caproto backend in ophyd get in here unfiltered? If so, we won't necessarily be limited to `PrimitiveType` - we may have those numpy types, arrays, etc.

atef/reduce.py
tangkong: I was thinking about the case where some aberrant enum (`['Out', '2', 'YAG']`), where both int and str "2" would be valid (so both types would be valid).  Do we think this is a fringe enough edge case to ignore?

atef/check.py
tangkong: It looks like we have a default reduction method (average) everywhere, and our data collection would methods would raise if the method were missing.

It was also my understanding that we don't really enforce PrimitiveType inside the code, but trying to deserialize a `Value` that holds a non-primitive type will fail. I think it's fair to say we don't support non-primitive values yet ([issue](https://github.com/pcdshub/atef/issues/160))

atef/reduce.py
tangkong: The answer I've come to is: yes probably

ZLLentz (approved):
I still think this looks good

tangkong:
Alright, I'll merge this for the same of moving on.  Thanks for the reviews 👍

PR #221, opened 2023-10-24
ENH: Improve status output

### LINKED ISSUE #146: GUI: add ✅ ❌  icons to run mode tree-view
## Expected Behavior
A clear icon would help greatly in quickly scanning to see what steps passed or failed

image

## Current Behavior
Nothin graphical in the tree view

## Possible Solution
Nothing too specific yet, but it's certainly doable.... right?

## Context
Brought up by Zach in https://github.com/pcdshub/atef/pull/140, but has been mentioned by almost everyone (Ken, Alex, etc)

## Your Environment
pcds-5.6.0

tangkong:
## Description
- Refactors "walk" helpers to a separate submodule, because I was getting confused and caught myself re-inventing the wheel
- Use ConfigTreeModel to show results in the tree view during run-mode (closes #146)
- Adds a "Results Summary" widget that shows a filterable view of checkout results, as well as a plaintext csv view for copying etc

## Motivation and Context
Also jira [ECS-1380](https://jira.slac.stanford.edu/browse/ECS-1380?jql=project%20%3D%20ECS%20AND%20component%20%3D%20ATEF)

Because the next thing I intend to do is rework how the page loading is done (for performance reasons), I expect the way trees are constructed to change significantly.  There's a couple places where I left todos to circle back on after I've settled on an approach.  

## How Has This Been Tested?
Interactively, a couple tests added

This PR

image

image

## Pre-merge checklist
- [x] (🤮)Test suite passes on GitHub Actions (😭)

### DIFF
--- a/atef/procedure.py
+++ b/atef/procedure.py
@@ -58,26 +58,6 @@ BS_STATE_MAP: Dict[int, BlueskyState] = {}
 MAX_PLAN_DEPTH = 200
 
 
-def walk_steps(
-    step: Union[ProcedureStep, PreparedProcedureStep]
-) -> Generator[Union[ProcedureStep, PreparedProcedureStep], None, None]:
-    """
-    Yield ProedureSteps in ``step``, depth-first.
-
-    Parameters
-    ----------
-    step : ProcedureStep
-        Step to yield ProcedureSteps from
-
-    Yields
-    ------
-    Generator[ProcedureStep, None, None]
-    """
-    yield step
-    for sub_step in getattr(step, 'steps', []):
-        yield from walk_steps(sub_step)
-
-
 @dataclasses.dataclass
 @serialization.as_tagged_union
 class ProcedureStep:
@@ -658,6 +638,13 @@ class PreparedProcedureGroup(PreparedProcedureStep):
         self.step_result = result
 
         return super().result
+
+    def walk_steps(self) -> Generator[AnyPreparedProcedure]:
+        for step in self.steps:
+            step = cast(AnyPreparedProcedure, step)
+            yield step
+            if hasattr(step, 'walk_steps'):
+                yield from step.walk_steps()
 
 
 @dataclass
--- a/atef/procedure.py
+++ b/atef/procedure.py
@@ -58,7 +58,9 @@ BS_STATE_MAP: Dict[int, BlueskyState] = {}
 MAX_PLAN_DEPTH = 200
 
 
-def walk_steps(step: ProcedureStep) -> Generator[ProcedureStep, None, None]:
+def walk_steps(
+    step: Union[ProcedureStep, PreparedProcedureStep]
+) -> Generator[Union[ProcedureStep, PreparedProcedureStep], None, None]:
     """
     Yield ProedureSteps in ``step``, depth-first.
 
--- a/atef/walk.py
+++ b/atef/walk.py
@@ -1,0 +1,175 @@
+"""
+Helpers for walking dataclasses.
+
+If relying on Prepared dataclass walk methods (walk_groups, walk_comparisons),
+note that if a configuration fails to prepare, they will be skipped
+"""
+from __future__ import annotations
+
+from typing import Any, Generator, List, Tuple, Union
+
+from atef.check import Comparison
+from atef.config import (Configuration, PreparedComparison,
+                         PreparedConfiguration, PreparedFile)
+from atef.procedure import (PreparedProcedureFile, PreparedProcedureStep,
+                            ProcedureStep)
+
+
+def walk_config_file(
+    config: Union[PreparedFile, PreparedConfiguration, PreparedComparison],
+    level: int = 0
+) -> Generator[Tuple[Any, int], None, None]:
+    """
+    Yields each config and comparison and its depth
+    Performs a recursive depth-first search
+
+    Parameters
+    ----------
+    config : Union[PreparedFile, PreparedConfiguration, PreparedComparison]
+        the configuration or comparison to walk
+    level : int, optional
+        the current recursion depth, by default 0
+
+    Yields
+    ------
+    Generator[Tuple[Any, int], None, None]
+    """
+    yield config, level
+    if isinstance(config, PreparedFile):
+        yield from walk_config_file(config.root, level=level+1)
+    elif isinstance(config, PreparedConfiguration):
+        if hasattr(config, 'configs'):
+            for conf in config.configs:
+                yield from walk_config_file(conf, level=level+1)
+        if hasattr(config, 'comparisons'):
+            for comp in config.comparisons:
+                yield from walk_config_file(comp, level=level+1)
+
+
+def walk_procedure_file(
+    config: Union[PreparedProcedureFile, PreparedProcedureStep, PreparedComparison],
+    level: int = 0
+) -> Generator[Tuple[Any, int], None, None]:
+    """
+    Yields each ProcedureStep / Comparison and its depth
+    Performs a recursive depth-first search
+
+    Parameters
+    ----------
+    config : Union[PreparedProcedureFile, PreparedProcedureStep,
+                    PreparedComparison]
+        the item to yield and walk through
+    level : int, optional
+        the current recursion depth, by default 0
+
+    Yields
+    ------
+    Generator[Tuple[Any, int], None, None]
+    """
+    yield config, level
+    if isinstance(config, PreparedProcedureFile):
+        yield from walk_procedure_file(config.root, level=level+1)
+    elif isinstance(config, PreparedProcedureStep):
+        for sub_step in getattr(config, 'steps', []):
+            yield from walk_procedure_file(sub_step, level=level+1)
+        if hasattr(config, 'walk_comparisons'):
+            for sub_comp in config.walk_comparisons():
+                yield from walk_procedure_file(sub_comp, level=level+1)
+
+
+def walk_steps(
+    step: Union[ProcedureStep, PreparedProcedureStep]
+) -> Generator[Union[ProcedureStep, PreparedProcedureStep], None, None]:
+    """
+    Yield ProedureSteps in ``step``, depth-first.
+
+    Parameters
+    ----------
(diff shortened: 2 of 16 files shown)


tangkong:
The test suite has been immensely irritating, passes normally and failing randomly on GHA.  I'm going to ignore it for the most part and circle back in yet another test suite fixup effort.  (when will it end...)

tangkong:
Of course the tests pass first try now. 🥲   I've punted a good amount of stuff here (including test cleanup pt 2) until after reworking the page-widget-loading system, but everything here should stand on its own I hope.

docs/source/upcoming_release_notes/221-enh_status_out.rst
```diff
@@ -0,0 +1,23 @@
+221 enh_status_out
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- Adds results summary page accessible from run-mode
```
klauer, line 10: Adding "in the `atef config` GUI" for future disambiguation could be a good thing, even if it's understood now

atef/walk.py
```diff
@@ -0,0 +1,175 @@
...
+                yield from walk_config_file(conf, level=level+1)
+        if hasattr(config, 'comparisons'):
+            for comp in config.comparisons:
+                yield from walk_config_file(comp, level=level+1)
+
+
+def walk_procedure_file(
+    config: Union[PreparedProcedureFile, PreparedProcedureStep, PreparedComparison],
+    level: int = 0
+) -> Generator[Tuple[Any, int], None, None]:
```
klauer, line 52: Careful with the `Any` annotation - it specifically says "don't bother checking the type for this" and it's actually a bit worse than some alternatives ([ref](https://stackoverflow.com/questions/39817081/typing-any-vs-object))

Might be good to make unions for possible things to come out of this.

atef/widgets/config/result_summary.py
```diff
@@ -0,0 +1,497 @@
...
+    """ Normalized, and slightly processed view of configs/steps with results """
+    status: Severity
+    name: str
+    reason: str
+
+    # An un-prepared dataclass, to match the tree views (will not hold the result)
+    origin: AnyDataclass
+
+    @property
+    def type(self):
```
klauer, line 33:
```suggestion
    def type(self) -> str:
        """The class name of the origin."""
```

atef/widgets/config/result_summary.py
```diff
@@ -0,0 +1,497 @@
...
+        # check mark
+        Severity.success: ('\u2713', QtGui.QColor(0, 128, 0, 255)),
+        Severity.warning : ('?', QtGui.QColor(255, 165, 0, 255)),
+        # x mark
+        Severity.internal_error: ('\u2718', QtGui.QColor(255, 0, 0, 255)),
+        Severity.error: ('\u2718', QtGui.QColor(255, 0, 0, 255)),
+        'N/A': ('nothing', QtGui.QColor())
+    }
+
+    def __init__(self, *args, data: List[ResultInfo] = [], **kwargs) -> None:
```
klauer, line 53: Mutable in default kwargs - could lead to hard to debug problems

`Optional[List[ResultInfo]] = None` could suffice and then `self.result_info = data or []`

atef/widgets/config/result_summary.py
```diff
@@ -0,0 +1,497 @@
...
+        super().__init__(*args, **kwargs)
+        self.result_info = data
+        self.headers = ['Status', 'Type', 'Name', 'Reason']
+
+    @classmethod
+    def from_file(cls, file) -> ResultModel:
```
klauer, line 59: File type hint?

atef/widgets/config/result_summary.py
```diff
@@ -0,0 +1,497 @@
...
+from atef.walk import walk_config_file, walk_procedure_file
+from atef.widgets.core import DesignerDisplay
+from atef.widgets.utils import insert_widget
+
+
+@dataclasses.dataclass
+class ResultInfo:
+    """ Normalized, and slightly processed view of configs/steps with results """
+    status: Severity
+    name: str
```
klauer, line 26: `name` always seems to come from `origin` directly in the `from_file` classmethod below - could it be a property?

atef/widgets/config/utils.py
```diff
@@ -1159,7 +1160,17 @@ def data(self, index: QtCore.QModelIndex, role: int) -> Any:
...
+    def data_updated(self) -> None:
+        """ method for calling dataChanged on the entire view """
+        top_left = self.index(0, 0, QtCore.QModelIndex())
+        bottom_right = self.index(self.rowCount(QtCore.QModelIndex()),
+                                  self.columnCount(QtCore.QModelIndex()),
+                                  QtCore.QModelIndex())
+        self.dataChanged.emit(top_left, bottom_right)
 
+
+# TODO: Rename this and related helpers to be more specific
```
klauer, line 1172: TODO reminder; going to make an issue I assume?

atef/widgets/config/window.py
```diff
@@ -581,6 +583,91 @@ def __init__(
...
+        )
+        self.model = ConfigTreeModel(data=self.root_item)
+
+        self.tree_view.setModel(self.model)
+
+        self.tree_view.resizeColumnToContents(1)
+        self.tree_view.header().swapSections(0, 1)
+        self.tree_view.expandAll()
+        # Connect to pages
+        # TODO: when we redo tree construction, this may need looking at
```
klauer, line 602: TODO reminder

klauer (commented):
No major issues looking at the code - will try it out and report back

atef/widgets/config/utils.py
tangkong: Yeah I intend to address this and other TODOs when I rework widget / tree loading

klauer:
The results summary is awesome!

A few thoughts which you could file under "nice to haves" as they are not showstoppers and reflect my opinion only:
* Perhaps default to "show warning + error". Success steps are nice and all but I'm probably not looking at the dialog for that.
* Plain text is more like "CSV" which is nice
   * Do you make sure there are no commas in reasons to make importing into Excel a pain?
   * Consider maybe adding a "copy to clipboard" option there
   * It may also be sufficient to have a menu option in the tab "Table" that says "Copy results to clipboard as CSV" or "Save results to file as CSV" 
    * A `rich`-like tree text output that you get from running it in the console would be neat to have but 3 different views of the same result is probably stretching it
    * A "go to config item" if you double click on the table would be nice (or some other context menu)
* Some more keyboard shortcuts (in the vein of #223 ) would be nice
    * I'd like to be able to run the selected check clicking a bit less (F5 to run or something)
* It'd be nice if the config tree loaded up expanded (so you don't just see "Root" and have to remember to click the arrow next to it)

atef/widgets/config/result_summary.py
ZLLentz, line 1: Is there any way to track/force where the result summary window opens? For me it opened on a different monitor, and I suspect in the hutch they will have similar behavior in their 3+ monitor setups.

atef/procedure.py
ZLLentz, line 1: Is there something fundamentally broken with comparisons right now? (I only noticed because the new UI is clear) (it seems like this happens prior to this PR too)
```
    ├── ❌Error: Device
    │   ├── ✔: im1l0.focus_motor.user_readback <= 0.0: Warn if the focus motor is negative, indicating it should be homed: value of 76.184
    │   ├── ✔: im1l0.zoom_motor.user_readback < 1.0: Warn if the zoom motor is negative, indicating it should be homed: value of 30.012
    │   ├── ✔: im2l0.focus_motor.user_readback <= 0.0: Warn if the focus motor is negative, indicating it should be homed: value of 50.088
    │   ├── ✔: im2l0.zoom_motor.user_readback < 1.0: Warn if the zoom motor is negative, indicating it should be homed: value of 24.744
    │   ├── ✔: im4l0.focus_motor.user_readback <= 0.0: Warn if the focus motor is negative, indicating it should be homed: value of 35.16
    │   ├── ✔: im4l0.zoom_motor.user_readback < 1.0: Warn if the zoom motor is negative, indicating it should be homed: value of 25.245
```

image

ZLLentz (commented):
I like the updated UI elements a lot, I don't immediately have specific suggestions about them.
I don't have any specific recommendations in the code either, it looks pretty clean.

I saw some strange behavior in `atef check` that predates this PR
I had one question/suggestion re: opening windows in a multi-screen environment

atef/procedure.py
ZLLentz: OH, actually, the issue here is that the intent and the implementation are backwards in the example config: the check is supposed to be the "good" state, but this checks for the "bad" state

atef/widgets/config/result_summary.py
tangkong: I fixed this as I was addressing some of Ken's comments.  Great minds think alike

tangkong:
Thanks for the suggestions!  I've chosen to punt a few of them, which I'll justify below:

It may also be sufficient to have a menu option in the tab "Table" that says "Copy results to clipboard as CSV" or "Save results to file as CSV"
> A rich-like tree text output that you get from running it in the console would be neat to have but 3 different views of the same result is probably stretching it
  * Running `atef check` on active checkouts is broken because I so strongly believed users would want human interaction to verify steps.  Maybe once I get active checkouts running from a cli this could replace the `plain text` tab entirely
> A "go to config item" if you double click on the table would be nice (or some other context menu)
  * Punting on the off chance that whatever i implement here has to be rewritten when I muck around with the trees.  I'll definitely hit this in that effort though
> Some more keyboard shortcuts (in the vein of https://github.com/pcdshub/atef/issues/223 ) would be nice
I'd like to be able to run the selected check clicking a bit less (F5 to run or something)
  * I'll tackle this in a separate PR, I think this is a good idea

tangkong:
Any further thoughts on this?  Or shall I merge and move on?

ZLLentz (approved):
No further thoughts- I like this a lot, it improves the UX substantially

klauer (approved):

Issue #222, opened 2023-10-26
PERF / GUI: Large checkouts take prohibitively long to open


tangkong:
## Expected Behavior
We should be able to open any size checkout

## Current Behavior
LAS made quite a large checkout that takes forever to open in the config GUI.  The checkout can still be run from the command line via `atef check`, but that's not an adequate solution.

checkout path: /cds/home/r/rj1/atef/timing_config/NEH_FS14/

## Possible Solution
Figure out a way to make widget loading more lazy.  Some more debug logging would be helpful here too

## Steps to Reproduce (for bugs)
1. atef --log DEBUG config /cds/home/r/rj1/atef/timing_config/NEH_FS14/

## Your Environment
pcds-5.8.0, atef-1.2.0

tangkong:
Important thought: Tree-creation performed when the widgets are loaded.  For example, A configuration group will make AtefItem's and [attach them to parents when a row is added](https://github.com/pcdshub/atef/blob/a775d1a99c2ebdbc9192daa404ac4029fa7e3105/atef/widgets/config/page.py#L827) (whether from loading a file or gui intereaction).  

In order to load widgets lazily, we'll have to at least decouple the tree creation from widget creation.  

I think ideally we have a system where the tree is created initially, and widgets are created once the tree-item is accessed.  

I wonder if this also somehow allows for widgets to be garbage collected.  (what's the memory load for those 3000 comparisons?)  Maybe naively we have some queue storing pages with a size limit.  

-----
Zach has been looking at improving pmps gui performance, and has taken the approach of treating widget-loads as coroutines to be kicked off in series. ([eg](https://github.com/pcdshub/pmps-ui/blob/bf437554acc40b399aa407b2508d6d1be969c2a3/templates/arbiter_outputs_entry.py#L44-L46)) Critically this means the surrounding tab can be loaded immediately and the contained tables will be populated gradually. 

While I don't think this is immediately applicable here, I think it's good reference.  In the atef case, a user might want to access the last item in the tree first.  In a large, roughly flat checkout that might not load until the very end.

tangkong:
With a combination of #224 and #226, this has been resolved.  Testing was done on the problematically huge config, and loading is much quicker now.

PR #224, opened 2023-11-10
PERF/REF: Lazy Page loading

tangkong:
## Description
This is going to be a doozy.  This has the simple goal of lazily loading `PageWidgets`, and the not-so-simple reality of rewriting most of said widgets.

A high-level list of things I did
- Consolidate `EditTree` and `RunTree` into a single `DualTree`
- Replace `AtefItem` (QTreeWidgetItem) with `TreeItem, for use in a `QTreeView` with `ConfigTreeModel` (a tree model with status icons)
- Moved "tree logic" into the dataclasses, out of the `PageWidgets`
- Moved `QDataclassBridge` instances to the `TreeItem`
- **Implemented a widget cache, and set up `DualTree` to load widgets lazily**

## Motivation and Context
(#222)
One user programmatically generated a checkout with 3000 comparisons, and the GUI failed to load it.  It attempted to, but way ATEF is structured every single widget had to be loaded.  

Briefly, the previous functionality was:
- **Tree creation logic was tied to widgets.**
- Opening a checkout only loaded the root node.  The construction of the root widget then created its children tree items and widgets, and those children loaded their own children, etc.
- This also means we have had to simulate the logic stored in the widgets if we wanted to re-create the checkout tree outside of widgets (e.g. reports, run-mode-tree)

To accomplish this we have to extricate the tree logic from the widgets.
- This means the dataclasses understand how to yield pieces the of their data that get pages in the gui, similar to how they know how to yield their comparisons etc.
- This prevents us from having to re-write tree-walking helpers that replicate   tree construction, consolidation yay!

A key requirement is that each widget must be loadable at any point by selecting its tree item.
- Because widgets were created by their parents, they grew to rely on methods from their parents
- All widgets can now be constructed given only their tree item and a DualTree instance, which holds the File-dataclass the data lives in.
  - Rather than reaching for the parent widget, a widget will reach for its parent `TreeItem`.  
- Callbacks that update the parent widget will now only happen if that widget exists.  If the widget doesn't exist, it will be created (and already updated) when requested

Because widgets always existed, it was natural to tie the QDataclassBridges to the widgets.  As such any update behavior/callbacks need to access widgets (that now may not exist).  I think it makes the most sense here to move the bridges to the `TreeItem`, and base behavior on that.  (this was quite disruptive)
- `PageWidget`s now need the tree at __init__-time (compared to the previous additional `assign_tree_items` step
- This simplifies the writing of a Pagewidget, since now we don't need to remember to call `assign_tree_items` outside of `__init__`.  I've also changed the name to more accurately reflect its use.
- Widgets need to notify the model if they make any changes to the underlying data structure, but can do so since they hold their `DualTree`

## How Has This Been Tested?
Test suite runs, but mostly interactively.   Perhaps I need to expand the test suite more...

Verbosely in this PR, as I always do

## Pre-merge checklist


atef/bin/config.py
```diff
@@ -58,6 +58,13 @@ def build_arg_parser(argparser=None):
     https://doc.qt.io/qt-5/qapplication.html#QApplication
     https://doc.qt.io/qt-5/qguiapplication.html#supported-command-line-options
     """
+    argparser.add_argument(
+        "--cache_size",
```
klauer, line 62:
```suggestion
        "--cache-size",
```
My opinion - mixing of `-` and `_` is annoying in CLI args

atef/tests/conftest.py
```diff
@@ -330,11 +330,17 @@ def configuration_group():
...
-        cfg_page = page_cls(cfg)
-        cfg_item = AtefItem(QtWidgets.QTreeWidget(), name='root',
-                            func_name=page_cls.__name__)
-        link_page(item=cfg_item, widget=cfg_page)
+        if isinstance(cfg, get_args(AnyConfiguration)):
+            file = ConfigurationFile()
+            file.root.configs.append(cfg)
+        elif isinstance(cfg, get_args(AnyProcedure)):
+            file = ProcedureFile()
+            file.root.steps.append(cfg)
```
klauer, line 338: For future purposes:

`else: raise NotImplementedError()`

atef/widgets/config/page.py
```diff
@@ -410,44 +321,42 @@ def setup_parent_button(self, button: QToolButton) -> None:
...
+    def setup_child_button(self, button: QToolButton, item: TreeItem) -> None:
         """
         Set up a button's style and make it navigate to a specific child page.
         """
-        # Create a function that navigates to the right page
-        def inner_navigate(*args, **kwargs):
-            self.navigate_to(item)
+        navigate_slot = WeakPartialMethodSlot(button, button.clicked,
+                                              self.navigate_to, item)
+        self._partial_slots.append(navigate_slot)
```
klauer, line 330: If we expect there to be a lot of these, we might consider pruning the list occasionally (i.e., removing those that have been unsubscribed)

atef/widgets/config/page.py
```diff
@@ -2207,14 +2100,41 @@ def select_comparison_type(self, new_type_index: int) -> None:
...
-            new_comparison=comparison,
-            comp_item=self.tree_item,
-        )
-        self.tree_item.setText(1, new_type.__name__)
-        self.new_comparison(comparison=comparison)
-        self.update_context()
+        parent_widget = self.full_tree.maybe_get_widget(self.parent_tree_item)
+
+        with self.full_tree.modifies_tree(select_prev=False):
+            # TODO: Still need to replace in dataclass.
```
klauer, line 2106: TODO reminder (how important is this to end user?)

atef/widgets/config/utils.py
```diff
@@ -677,7 +677,30 @@ def update_widget(value: Any) -> None:
     value_obj.changed_value.connect(update_widget)
 
 
-def describe_comparison_context(attr: str, config: Configuration) -> str:
+def get_comp_field_in_parent(
+    comp: Comparison,
+    parent: Union[Configuration, ProcedureStep]
+) -> str:
```
klauer, line 683: Docstring?

atef/widgets/config/utils.py
```diff
@@ -1264,6 +1321,10 @@ def child(self, row: int) -> TreeItem:
         if row >= 0 and row < self.childCount():
             return self._children[row]
 
+    def get_children(self) -> Generator[TreeItem, None, None]:
+        """ Yield this item's children """
```
klauer, line 1325: This one may make you yell at your monitor (sorry!) - style guide does say "no extra whitespace with a single line docstring" [ref](https://confluence.slac.stanford.edu/display/PCDS/ECS+Python+Standards+and+Style+Guide#ECSPythonStandardsandStyleGuide-SinglelineDocstrings)

It's another thing we should be linting for at some point

atef/widgets/config/utils.py
```diff
@@ -1284,7 +1345,77 @@ def addChild(self, child: TreeItem) -> None:
         child._parent = self
         child._row = len(self._children)
         self._children.append(child)
-        self._columncount = max(child.columnCount(), self._columncount)
+
+    def removeChild(self, child: TreeItem) -> None:
```
klauer, line 1349: Where did the camelcase come in? Mimicking qt api?

klauer (commented):
There's a lot to process here... Overall:
* I think this is on the right track separating out logic from the GUI a bit more
* It seemed to work well interactively aside from the one crash 👍

klauer:
Crash:

<details>

```
$ python -m atef config atef/tests/configs/active_test.json
(click run; then click through the tree - and it crashes at SetValueStep)
QThread: Destroyed while thread is still running
Abort trap: 6
(atef) PC98125:atef klauer$ /Users/klauer/mc/envs/atef/lib/python3.9/multiprocessing/resource_tracker.py:216: UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```
Repeatable on psbuild so I used gdb there:
```
(gdb) py-bt
Traceback (most recent call first):
  File "/cds/home/k/klauer/Repos/atef/atef/widgets/config/window.py", line 626, in show_page_for_data
    new_widget = self.create_widget(item, mode)
  File "/cds/home/k/klauer/Repos/atef/atef/widgets/config/window.py", line 610, in show_selected_display
```
`thread apply all py-bt` marked this thread sitting off in libca land:
```
Thread 56 (Thread 0x7fff78a80700 (LWP 18193) "BusyCursorThrea"):
Traceback (most recent call first):
  Waiting for the GIL
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.8.0/lib/python3.9/site-packages/epics/ca.py", line 910, in pend_event
    ret = libca.ca_pend_event(timeout)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.8.0/lib/python3.9/site-packages/epics/ca.py", line 923, in poll
    pend_event(evt)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.8.0/lib/python3.9/site-packages/epics/ca.py", line 547, in wrapper
    return fcn(*args, **kwds)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.8.0/lib/python3.9/site-packages/epics/ca.py", line 1072, in connect_channel
    poll()
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.8.0/lib/python3.9/site-packages/epics/ca.py", line 577, in wrapper
    return fcn(*args, **kwds)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.8.0/lib/python3.9/site-packages/epics/pv.py", line 440, in connect
    ca.connect_channel(self.chid, timeout=timeout)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.8.0/lib/python3.9/site-packages/epics/pv.py", line 47, in wrapped
    return func(self, *args, **kwargs)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.8.0/lib/python3.9/site-packages/epics/pv.py", line 425, in wait_for_connection
    self.connect(timeout=timeout)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.8.0/lib/python3.9/site-packages/epics/pv.py", line 54, in wrapped
    return func(self, *args, **kwargs)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.8.0/lib/python3.9/site-packages/ophyd/signal.py", line 1229, in _ensure_connected
    if not pv.wait_for_connection(timeout=timeout):
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.8.0/lib/python3.9/site-packages/ophyd/signal.py", line 1786, in wait_for_connection
    self._ensure_connected(self._read_pv, self._write_pv, timeout=timeout)
  File "/cds/home/k/klauer/Repos/atef/atef/widgets/config/data_active.py", line 937, in get_curr_value
    self._sig.wait_for_connection()
  File "/cds/home/k/klauer/Repos/atef/atef/widgets/utils.py", line 202, in run
    self.func()
```
</details>

tangkong:
RE: crash

I went through earlier and replaced functions-defined-in-functions with WeakPartialMethodSlot, but didn't do the same for functions passed to QThreads.  I wonder if making those functions proper methods would help there.

atef/widgets/config/window.py
```diff
@@ -455,379 +455,305 @@ def setup_ui(self):
...
+            # not in cache, need to build the widget.
+            new_widget = self.create_widget(item, mode)
 
-        if item is None:
-            logger.error('Tree has no currentItem.  Cannot show selected')
-            return
-        if item is self.last_selection:
-            return
+            if len(curr_cache) >= self.max_cache_size:
+                _, oldest_widget = curr_cache.popitem(last=False)
```
klauer, line 637: I worry a bit about this caching mechanism causing unexpected side effects (orphaned widgets, crashes, etc) - but admittedly I'm not the most familiar with the inner workings here. 

Mind commenting on the intention? I assume the concern is memory usage when someone clicks around on a bunch of widgets and loads them all up?

atef/config.py
```diff
@@ -95,6 +145,35 @@ class DeviceConfiguration(Configuration):
...
+        Replace ``old_comp`` with ``new_comp`` in this dataclass, wherever it is.
+        Looks through ``shared``, then any of the attribues in ``comp_attrs``
+
+        Parameters
+        ----------
+        old_comp : Comparison
+            Comparison to be replaced
+        new_comp : Comparison
+            Comparison to replace ``old_comp`` with
+        comp_attrs : Optional[List[str]], optional
```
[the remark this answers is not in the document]
tangkong: I thought so too, but my auto-docstring generator kept putting it in and I stopped fighting it.  I wonder if there's a way to change this in my extension

atef/config.py
```diff
@@ -312,6 +453,10 @@ def walk_groups(
         yield self.root
         yield from self.root.walk_groups()
 
+    def children(self) -> List[PreparedGroup]:
```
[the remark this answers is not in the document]
tangkong: I end up never calling `.children()` on `PreparedFile` so this ends up not being necessary.  It is indeed a nearly identical method

atef/procedure.py
```diff
@@ -98,6 +103,18 @@ def walk_steps(self) -> Generator[AnyProcedure, None, None]:
...
 
+    def children(self) -> List[Union[ProcedureStep, ProcedureGroup]]:
+        """ Return children of this group, as a tree view might expect """
+        return self.steps
+
+    def replace_step(
+        self,
+        old_step: Union[ProcedureStep, ProcedureGroup],
+        new_step: Union[ProcedureStep, ProcedureGroup]
+    ) -> ProcedureStep:
```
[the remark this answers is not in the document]
tangkong: A vestigial return type that escaped the purge 😬

atef/widgets/config/page.py
```diff
@@ -788,13 +706,20 @@ def add_config_row(
...
             self.data.configs.append(config)
         config_row = ConfigurationGroupRowWidget(data=config)
-        config_page = PAGE_MAP[type(config)](data=config)
-        config_item = AtefItem(
-            tree_parent=self.tree_item,
-            name=config.name or 'untitled',
-            func_name=type(config).__name__,
-        )
-        link_page(item=config_item, widget=config_page)
+        # TODO Still broken, no child setup row, no init setup
```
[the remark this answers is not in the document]
tangkong: I think I fixed this actually, whoops

atef/widgets/config/page.py
tangkong: Also fixed 🤦

atef/widgets/config/utils.py
tangkong: I'll fix this for the entire repo, at the expense of the diff

atef/widgets/config/utils.py
tangkong: Yeah, this was done in an effort to mimic QTreeWidgetItem and smooth the transition between `AtefItem` and `TreeItem`, though I ended up mixing styles at some point.  I think at this point we can probably convert everything to snake_case, since this is sufficiently hidden behind the `ConfigTreeModel`

atef/widgets/config/window.py
tangkong: Memory usage was the concern, since it's easy to just hold the down-arrow and scroll through all the widgets.  (or spam the next-button in run-mode)

I haven't run into issues yet, and the shift to making each PageWidget self sufficient is important to enabling this.  (as in a PageWidget will never ask for methods on other PageWidgets).  If a PageWidget is deleted, nothing should ever want to access it.  

The concerns are valid though, I can't confidently say there will be no crashes.  (as your previous comment shows)

atef/widgets/config/page.py
klauer: Or perhaps better: `WeakPartialMethodSlot` could become a `QObject`, emit a "Destroyed!" signal when any goes out of scope - then `_partial_slots` could be cleaned automatically... Hmmm

atef/widgets/config/data_base.py
```diff
@@ -45,6 +47,9 @@ class DataWidget(QWidget):
         Even parent is unlikely to see use because parent is set automatically
         when a widget is inserted into a layout.
     """
+    # TODO: Remove this from DataWidget.  Bride should live in TreeItem to
```
ZLLentz, line 50:
```suggestion
    # TODO: Remove this from DataWidget.  Bridge should live in TreeItem to
```

atef/config.py
```diff
@@ -52,6 +52,53 @@ class Configuration:
     #: Tags tied to this configuration.
     tags: Optional[List[str]] = None
 
+    def children(self) -> List[Any]:
+        """ Return children of this group, as a tree view might expect """
```
ZLLentz, line 56: nitpick: lingering spacy comments in this file

atef/procedure.py
```diff
@@ -84,6 +85,10 @@ def allow_verify(self) -> bool:
         """
         return self.result.severity == Severity.success
 
+    def children(self) -> List[Any]:
+        """ Return children of this group, as a tree view might expect """
```
ZLLentz, line 89: nitpick: lingering spacy comments in this file

atef/widgets/config/page.py
```diff
@@ -324,50 +235,50 @@ class PageWidget(QWidget):
...
+        data: AnyDataclass,
+        tree_item: Optional[TreeItem] = None,
+        full_tree: Optional[DualTree] = None,
+        **kwargs
+    ) -> None:
         super().__init__(**kwargs)
         self.data = data
+        if tree_item is not None:
+            # assign tree item and check
+            assert tree_item.orig_data is data
```
ZLLentz, line 249: Assert statements aren't necessarily run at runtime: https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement, so if this is important it should probably be an if/raise sort of construct

atef/widgets/config/window.py
ZLLentz: Do you know how much memory, say, ~1000 (hidden) page widgets uses? It would be useful to quantify this to understand what the real impacts are here. I personally suspect that crashing concerns on this PR and on any future expansions to atef are much more real than the RAM usage invested here, but without knowing the actual numbers it's impossible to confirm.

ZLLentz (commented):
This overall is a big improvement over the previous implementation, even outside the lazy loading.

There's too much here to give a timely review that checks every single line but I read a lot of them.

I'm going to proceed to interactive testing.

atef/check.py
```diff
@@ -843,3 +869,7 @@ async def prepare(self, cache: Optional[DataCache] = None) -> None:
             await self.warn_high_dynamic.prepare(cache)
             self.warn_high = self.warn_high_dynamic.get()
         self.is_prepared = True
+
+
+ALL_COMPARISONS = [Equals, NotEquals, Greater, GreaterOrEqual, Less, LessOrEqual,
+                   Range, ValueSet, AnyValue, AnyComparison]
```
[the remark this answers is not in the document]
klauer: Could technically be assembled automatically, if desirable:
```
>>> atef.check.Comparison.__subclasses__()
[<class 'atef.check.BasicDynamic'>, <class 'atef.check.ValueSet'>, <class 'atef.check.AnyValue'>, <class 'atef.check.AnyComparison'>, <class 'atef.check.Range'>]
```
Or some sort of registration system

But yeah, I think the "we're probably not going to add more" is reason enough to just keep this simple (i.e., as-is)

atef/config.py
tangkong: How did this escape the filter!?  (answer: my regex was bad)

atef/check.py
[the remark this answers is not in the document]
ZLLentz: Yeah, there are a few ways to do this, even auto-registration in init subclass for example
But maybe it's good to limit the scope of atef anyway and not open the comparison floodgates

atef/bin/config.py
ZLLentz, line 1: Another edge case exception unrelated to the crash Ken observed. This happened when I opened up the all fields example, clicked around a lot and stress tested changing values for a bit, switched to run mode, ran all the checks, switched back to config mode and started clicking around again.

image

ZLLentz: Interestingly, the crash observed for active_test doesn't seem to hit the big passive config here at all, allowing me to click around for a long time before hitting something like this

ZLLentz: It seems like the important thing here are:

- renaming the page
- switching to run mode and back to config mode

and the rest were just red herrings from me clicking around a lot

ZLLentz: And lastly this seems unrelated to the caching/freeing mechanisms (still happens with large cache size)

tangkong:
A realization about Ken's crash: the segfault is happening in an Edit widget, while SetValueStep has as unique run-widget.  The edit-widget should not be invoked.  (Past me left a TODO I see...)

ZLLentz:
(A little late: I've confirmed that this loads the user's big config quickly, solving the inciting incident)

atef/check.py
ZLLentz, line 1: RE: the user's big config (inciting incident):
- The gui opens quickly now :tada:
- There is a noticable freeze/slowdown when generating a PV configuration page that contains many many many comparisons

atef/widgets/config/page.py
tangkong: Removed this because it is actually wrong.  (PageWidgets can have prepared dataclasses as their `data`)

atef/bin/config.py
tangkong: > Interestingly, the crash observed for active_test doesn't seem to hit the big passive config here at all

Valuing clarity over redundancy, this crash happened because the edit widget was being created even if there was a unique run widget being built and used instead.  In passive checkouts, there are no unique run-widgets, so all the edit widgets are converted into run widgets and displayed, avoiding this case entirely.

tangkong: Zach's crash here seems to be a case where garbage collection deletes the widget before we can delete it ourselves.  Fixed in 9a149c5 by calling delete right after we pop it from the cache, but also catching RuntimeErrors out of an abundance of caution

tangkong:
After some "memory profiling" with top and the huge config, PVConfigurationPage's with many sub-comparisons still take prohibitively long to load, likely due to having to load all the row widgets in the comparison table.  This will be an issue for other ConfigurationGorup pages, and is seen in other apps like the pmps-ui.

External to the effort, investigating paginated table layouts is worthwhile

atef/widgets/config/window.py
ZLLentz: our server has 50.46 GB of RAM
opening the big file with no pages loaded is 0.8% memory usage = 0.4 GB consumed
with all the comparison pages loaded we get 3.5% usage -> 2.7% net = 1.362 GB RAM for just the pages
the config has 2541 Equals comparisons
so this means we eat up approx 536 kB per comparison page

I _think_ this means we're typically going to be OK because the example here is not reasonable or maintainable

klauer (approved):

ZLLentz (approved):

tangkong:
Thanks everyone.  Shipping this off

PR #226, opened 2023-12-04
ENH: Add and apply PagedTableWidget

tangkong:
## Description
Replaces the existing `QTableWidget` + `cellWidget` combination with a `PagedTableWidget`, for splitting up rows into pages.

## Motivation and Context
Previously, all the row widgets were created on PageWidget creation.  This can lead to long loading times that aren't remedied by the previous lazy page loading effort. 

This may seem a bit complicated, but should be the proper way to let qt handle construction and destruction of row widgets.  There was some complexity involved when hooking up buttons / slots inside these row widgets, since widgets can be created or destroyed at any time.  This means saving connection slots in a way that can be re-initialized at will.

I made an effort to keep the existing widgets and their functionality.

## How Has This Been Tested?
interactively

This PR

image

## Pre-merge checklist

### DIFF
--- a/atef/widgets/config/page.py
+++ b/atef/widgets/config/page.py
@@ -46,8 +46,7 @@ from atef.widgets.config.data_active import (CheckRowWidget,
                                              GeneralProcedureWidget,
                                              PassiveEditWidget,
                                              SetValueEditWidget)
-from atef.widgets.config.paged_table import (SETUP_SLOT_ROLE, USER_DATA_ROLE,
-                                             PagedTableWidget)
+from atef.widgets.config.paged_table import SETUP_SLOT_ROLE, PagedTableWidget
 from atef.widgets.config.run_active import (DescriptionRunWidget,
                                             PassiveRunWidget,
                                             SetValueRunWidget)
@@ -440,11 +439,6 @@ class PageWidget(QWidget):
 
         # Remove row from the table
         table.remove_data(row.data)
-        # for row_index in range(table.rowCount()):
-        #     widget = table.cellWidget(row_index, 0)
-        #     if widget is row:
-        #         table.removeRow(row_index)
-        #         break
         # Remove configuration from the data structure
         self.remove_table_data(data)
 
@@ -889,7 +883,7 @@ class DeviceConfigurationPage(DesignerDisplay, PageWidget):
                 )
 
         row_count = self.comparisons_table.row_count()
-        self.comparisons_table.insert_setup_row(
+        self.comparisons_table.insert_row(
             row_count,
             comparison,
             partial(self.configure_row_widget,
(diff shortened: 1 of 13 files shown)


atef/widgets/config/page.py
```diff
@@ -965,21 +1037,24 @@ class PVConfigurationPage(DesignerDisplay, PageWidget):
...
         self.insert_widget(
             self.pv_configuration_widget,
             self.pv_widget_placeholder,
         )
+
+        self.comparisons_table = PagedTableWidget(title="Comparisons")
+        self.insert_widget(
+            self.comparisons_table,
+            self.comparisons_table_placeholder
+        )
```
ZLLentz, line 1057: You should know that you can actually promote widgets to python widgets inside the ui file using designer, removing the need to have a replaceable placeholder widget. It seems to work here though so no real need to refactor. (I also forget the precise magic for doing this)

docs/source/upcoming_release_notes/226-enh_paged_table.rst
```diff
@@ -0,0 +1,22 @@
+226 enh_paged_table
+###################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- Adds PagedTableWidget, apply it to passive checkout group pages
```
ZLLentz, line 10:
```suggestion
- Adds PagedTableWidget, apply it to passive checkout group pages
  to substantially improve the loading performance of group pages with many, many comparisons.
```

ZLLentz (commented):
Everything here looks pretty solid
There's some repeated code in the page widgets that might be possible to deduplicate (not fully sure) but this isn't really an issue
There's a lot of code to go through so I might have missed something..

Tomorrow when systems are fully up I want to open with the example "slow loading" page as the highest-level stress test. I'll do some interactive testing then as well.

atef/widgets/config/page.py
tangkong: Oh interesting.  I'll poke around.

tangkong: Done, cleans up the python classes a bit 👍

ZLLentz:
Interactive testing:
- Everything loads super fast, even the crazy big user file. The pages in that file that previously didn't load at all now load with >100 pages in the paginated table in approx 1 second rather than "never"- you can feel the delay but it feels fitting for such a big load to have some sort of delay
- I thought I found a segfault but in actuality this (not so secretly) requires pcdsutils v0.14.1, which I did not have installed. It may be worth bumping the build constraints to >=0.14.1
- Navigating the table feels great here, and makes it possible to find the configs you want even when there are hundreds of them (and it would be even better if the names in this example file were simpler)
- Adding and removing seems to work smoothly from what I've tried
- Run mode still works (and still runs the big file super quick, even from the UI)

ZLLentz:
I guess, in summary, my only real suggestion is to bump the required version of pcdsutils

ZLLentz (approved):

PR #228, opened 2024-01-12
ENH: add pmgr-to-atef-check script

tangkong:
## Description
A simple script to grab a config from pmgr and write an atef check to verify the values

## Motivation and Context
https://jira.slac.stanford.edu/browse/ECS-1350

I'm honestly not sure if this is the minimal amount of information needed to grab a configuration, but this is similar to the pcdsdevices pmgr usage, and seems to work.

## How Has This Been Tested?
I ran the script, wrote a checkout, and ran it properly

This PR, and some docstring

## Pre-merge checklist

### DIFF
--- a/scripts/pmgr_check.py
+++ b/scripts/pmgr_check.py
@@ -1,0 +1,189 @@
+"""
+This script creates an atef check from a pmgr configuration.  The configuration will
+be converted into a PVConfiguration.  Note that default tolerances will be used for
+checks.
+
+An example invocation might be:
+"python scripts/pmgr_check.py cxi "KB1 DS SLIT LEF" CXI:KB1:MMS:13 test_pmgr_checkout.json"
+"""
+import argparse
+import json
+import logging
+from typing import Any, Dict
+
+import apischema
+from pmgr import pmgrAPI
+
+from atef.check import Equals
+from atef.config import ConfigurationFile, ConfigurationGroup, PVConfiguration
+
+DESCRIPTION = __doc__
+logger = logging.getLogger()
+
+
+def get_pv(prefix: str, key: str):
+    """
+    Parse key from pmgr configuration data dictionary.  Keys are of the form:
+    'FLD_ACCL' or 'FLD_BDST', denoting the suffixes to append to `prefix`.
+
+    Ignores unrecognized keys (keys without expected prefixes)
+
+    Parameters
+    ----------
+    prefix : str
+        the EPICS PV prefix
+    key : str
+        the key from a pmgr configuration data dictionary
+
+    Returns
+    -------
+    str
+        a fully qualified EPICS PV
+    """
+    if 'FLD_' in key:
+        suffix = key.removeprefix('FLD')
+    elif 'PV_' in key:
+        suffix = key.removeprefix('PV')
+    else:
+        logger.debug(f'Unrecognized key provided: {key}')
+        return
+
+    # general string fixing... ew
+    suffix_parts = suffix.split("__")
+    new_suffix_list = [":".join(substr.split('_')) for substr in suffix_parts]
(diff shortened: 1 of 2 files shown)


scripts/pmgr_check.py
```diff
@@ -0,0 +1,189 @@
...
+def get_cfg_data(hutch: str, config_name: str) -> Dict[str, Any]:
+    """
+    Get pmgr config data corresponding to ``config_name`` and ``hutch``
+
+    Parameters
+    ----------
+    hutch : str
+        the hutch name, e.g. 'cxi'
+    config_name : str
+        the pmgr config name, e.g. 'KB1 DS SLIT LEF'
```
ZLLentz, line 69: I'm far too amused that the real config here is LEF instead of LEFT

tangkong: None of the names in pmgr really makes sense to me, so I've just accepted it :/  There are character length limits but this is certaintly not hitting it...

scripts/pmgr_check.py
```diff
@@ -0,0 +1,189 @@
...
+        logger.debug(f'Unrecognized key provided: {key}')
+        return
+
+    # general string fixing... ew
+    suffix_parts = suffix.split("__")
+    new_suffix_list = [":".join(substr.split('_')) for substr in suffix_parts]
+    suffix = '_'.join(new_suffix_list)
+    suffix = ".".join(suffix.rsplit(":", 1))
+    pv = prefix + suffix
+    return pv
```
ZLLentz, line 57: This isn't quite correct, specifically the FW_MEANS and REV_MEANS PVs are separated by colons instead of dots, e.g. currently the following PVs are checked:
```
CXI:KB1:MMS:13.FW_MEANS
CXI:KB1:MMS:13.REV_MEANS
```
But the correct PVs are:
```
CXI:KB1:MMS:13:FW_MEANS
CXI:KB1:MMS:13:REV_MEANS
```

iirc the `FLD_` PVs should always use a `.` delimiter while the `PV_` PVs should always use a `:` delimiter but I'm not 100% on that

scripts/pmgr_check.py
```diff
@@ -0,0 +1,189 @@
...
+        type=str,
+        help="name of stored pmgr configuration, case and whitespace sensitive."
+             "e.g. 'KB1 DS SLIT LEF'",
+    )
+
+    parser.add_argument(
+        "prefix",
+        type=str,
+        help="EPICS PV Prefix, e.g. 'CXI:KB1:MMS:13'",
+    )
```
ZLLentz, line 141: There might be some additional utility in:

- Getting the values using the currently selected configuration for the prefix
- Allowing multiple prefix/config name pairs in one call

Not sure how much should be in scope for the script vs "import me and do the special thing you want"

ZLLentz (commented):
I spot checked a bunch of values and they seemed to match the values stored in pmgr, I noted two things to look at

scripts/pmgr_check.py
ZLLentz: All of the other slits use the full LEFT so this is probably just an unfortunate eternal typo

scripts/pmgr_check.py
tangkong: You are correct.  The `FLD_` PVs are the only ones that line 55 should be applied to.

scripts/pmgr_check.py
tangkong: I wanted to avoid replicating functionality found in the pmgr cli utilities, so I think I lean toward not implementing suggestion 1 here.  

Suggestion 2 might be a smidge confusing in this script, but since we're lacking a good way to combine checkouts via the gui it might be necessary.  I'll add #2

ZLLentz (approved):
This LGTM

PR #229, opened 2024-01-17
MNT: Refactor scripts into `atef scripts` subcommand

tangkong:
## Description
Refactors the scripts in the separate "scripts" folder into a sub-subcommand, so people can use them without knowing the exact path

## Motivation and Context
If someone wanted to use these scripts before, they'd have to do call the precise path:

```
python ./atef/scripts/converter_v0.py [args, kwargs]
```

This seems less than accessible, so I wanted to fold them into the atef CLI.  We now scan the scripts submodule, and each sub-sub-module gets its own `atef scripts` subsubcommand

<details><summary>Example CLI invocations</summary>
<p>

```bash
$ atef -h
usage: atef [-h] [--version] [--log LOG_LEVEL] {check,config,scripts} ...

`atef` is the top-level command for accessing various subcommands.

Try:

    $ atef check --help
    $ atef config --help
    $ atef scripts --help

positional arguments:
  {check,config,scripts}
                        Possible subcommands

optional arguments:
  -h, --help            show this help message and exit
  --version, -V         Show the atef version number and exit.
  --log LOG_LEVEL, -l LOG_LEVEL
                        Python logging level (e.g. DEBUG, INFO, WARNING)
```

```bash
$ atef scripts -h
usage: atef scripts [-h] {converter_v0,pmgr_check} ...

Runs atef related scripts. Pick a subcommand to run its script

positional arguments:
  {converter_v0,pmgr_check}
                        available script subcommands

optional arguments:
  -h, --help            show this help message and exit
```

```bash
$ atef scripts pmgr_check -h
usage: atef scripts pmgr_check [-h] [--names PMGR_NAMES [PMGR_NAMES ...]] [--prefixes PREFIXES [PREFIXES ...]] [--table TABLE_NAME] hutch filename

This script creates an atef check from a pmgr configuration.  The configuration will
be converted into a PVConfiguration.  Note that default tolerances will be used for
checks.

An example invocation might be:
python scripts/pmgr_check.py cxi test_pmgr_checkout.json --names "KB1 DS SLIT LEF" --prefix CXI:KB1:MMS:13

positional arguments:
  hutch                 name of hutch, e.g. 'cxi'
  filename              Output filepath

optional arguments:
  -h, --help            show this help message and exit
  --names PMGR_NAMES [PMGR_NAMES ...], -m PMGR_NAMES [PMGR_NAMES ...]
                        a list of stored pmgr configuration names, case and whitespace sensitive.e.g. 'KB1 DS SLIT LEF'.  Length must match --prefixes
  --prefixes PREFIXES [PREFIXES ...], -p PREFIXES [PREFIXES ...]
                        a list of EPICS PV prefixes, e.g. 'CXI:KB1:MMS:13'.  Length must match --names
  --table TABLE_NAME, -t TABLE_NAME
                        Table type, by default 'ims_motor'
```

</p>
</details> 

One caveat here is that doing all this requires we import the modules.  This was how we did this before, but I'm realizing the sluggishness of the CLI is likely largely due to all the importing we do.  If we wanted to defer these imports, we'd have to hard code the subcommands and lose the import error messages that have saved me multiple times in the past

## How Has This Been Tested?
Interactively, I need to re-learn how to pytest CLI stuff without click

This PR, as well as in the cli help text

## Pre-merge checklist

### DIFF
--- a/atef/bin/scripts.py
+++ b/atef/bin/scripts.py
@@ -1,8 +1,5 @@
 """
 `atef scripts` runs helper scripts.  Scripts may be added over time.
-
-Try:
-
 """
 
 import argparse
@@ -20,6 +17,7 @@ def gather_scripts() -> Dict[str, Tuple[Callable, Callable]]:
     """Gather scripts, one main function from each submodule"""
     # similar to main's _build_commands
     global DESCRIPTION
+    DESCRIPTION += "\nTry:\n"
     results = {}
     unavailable = []
 
--- a/atef/bin/scripts.py
+++ b/atef/bin/scripts.py
@@ -1,0 +1,70 @@
+"""
+`atef scripts` runs helper scripts.  Scripts may be added over time.
+
+Try:
+
+"""
+
+import argparse
+import importlib
+import logging
+from pkgutil import iter_modules
+from typing import Callable, Dict, Tuple
+
+logger = logging.getLogger(__name__)
+
+DESCRIPTION = __doc__
+
+
+def gather_scripts() -> Dict[str, Tuple[Callable, Callable]]:
+    """Gather scripts, one main function from each submodule"""
+    # similar to main's _build_commands
+    global DESCRIPTION
+    results = {}
+    unavailable = []
+
+    scripts_module = importlib.import_module("atef.scripts")
+    for sub_module in iter_modules(scripts_module.__path__):
+        module_name = sub_module.name
+        try:
+            module = importlib.import_module(f".{module_name}", "atef.scripts")
+        except Exception as ex:
+            unavailable.append((module_name, ex))
+        else:
+            results[module_name] = (module.build_arg_parser, module.main)
+            DESCRIPTION += f'\n    $ atef scripts {module_name} --help'
+
+    if unavailable:
+        DESCRIPTION += '\n\n'
+
+        for command, ex in unavailable:
+            DESCRIPTION += (
+                f'\nWARNING: "atef scripts {command}" is unavailable due to:'
+                f'\n\t{ex.__class__.__name__}: {ex}'
+            )
+
+    return results
+
+
+SCRIPTS = gather_scripts()
+
+
+def build_arg_parser(argparser=None):
+    if argparser is None:
+        argparser = argparse.ArgumentParser()
+
+    argparser.description = """
+    Runs atef related scripts.  Pick a subcommand to run its script
+    """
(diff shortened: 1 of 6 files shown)


tangkong:
my local tests have begun segfaulting again.  New year new atef suite struggles 😢

ZLLentz:
> One caveat here is that doing all this requires we import the modules. This was how we did this before, but I'm realizing the sluggishness of the CLI is likely largely due to all the importing we do.

I noticed this in other libraries, the only way to make the CLI feel responsive was to do some very unpythonic import practices to avoid pulling in anything that isn't used

atef/bin/scripts.py
```diff
@@ -0,0 +1,70 @@
+"""
+`atef scripts` runs helper scripts.  Scripts may be added over time.
+
+Try:
+
+"""
```
ZLLentz, line 6: try

ZLLentz (dismissed):
this is reasonable
one silly comment

atef/bin/scripts.py
tangkong: This was to **try** and get the following
```
$ atef scripts

`atef scripts` runs helper scripts.  Scripts may be added over time.

Try:

    $ atef scripts converter_v0 --help
    $ atef scripts pmgr_check --help
```

I got a bit dizzy messing around with all the help text, and didn't realize it doesn't show up for `atef scripts -h`

ZLLentz: You can accomplish the same without putting the extra word here in the global docstring, right? It could have equivalently been added during `gather_scripts`?

ZLLentz: Or, if you were trying to mutate the global `__doc__`, you can do that I think but you have to use `global __doc__` instead of `global DESCRIPTION`

ZLLentz: Either way, I think having the dangling Try literal here decreases readability

tangkong: Oh I see what you mean now.  Yea I can move that without much issue

tangkong:
I found that increasing the widget cache size helped stave off the segfaults locally.  I'm not sure why this was an issue, since the segfault was occurring after only 2 pages were selected (root -> group page).  Perhaps I was also adding the page to qtbot too early.  Qt testing is always an adventure

ZLLentz (approved):

PR #231, opened 2024-02-07
BUG: remove deleteLater() and setParent(None) calls

### LINKED ISSUE #230: BUG: segfaults when switching windows
## Expected Behavior
We shouldn't segfault ever...

## Current Behavior
When thumbing through the all_fields.json test config, I ran into segfaults.  This was narrowed down to the [deleteLater() block](https://github.com/pcdshub/atef/blob/e961f98bfd65f73eb0ec41bcc8431cb131f0fd4f/atef/widgets/config/window.py#L560-L561) where we remove PageWidgets from the cache.  Apparently now the garbage collection is grabbing widgets before I can call deleteLater and set the parent to None.

This definitely didn't happen before, I'm relatively certain we all manually filled the cache to verify I wasn't doing just this.  

## Possible Solution
Let the garbage collector do its work.  (remove the deleteLater calls)
There's no great way to explicitly ensure the widgets are deleted, we kind of just have to trust things if we do this.  Perhaps this isn't too bad, and will be enough to stave off memory leaks.  

Ultimately this does open the window to widgets piling up and eating up memory...

1.  Open `all_fields.json` config
2. Click on a checkout in the tree view
3. down arrow through the configs until the segfault happens

## Context
Browsing a config

## Your Environment
pcds-5.8.1

tangkong:
## Description
Remove `deleteLater()` and `setParent(None)` calls from cached widget deletion process, as they seem to cause segfaults and are not strictly necessary...

## Motivation and Context
closes #230 

## How Has This Been Tested?
Interactively, I need to add (better) tests though (for a followup)

My interactive testing focused on making sure widgets were actually being deleted by garbage collection, despite not explicitly requesting it.  To do this I held a weakref to widgets that were popped from the cache, and observed that these weakrefs weren't piling up as I loaded and popped widgets.   This could be made more empirical of course, I'll hopefully fix that soon.

This PR

## Pre-merge checklist

### DIFF
--- a/atef/widgets/config/window.py
+++ b/atef/widgets/config/window.py
@@ -652,15 +652,6 @@ class DualTree(DesignerDisplay, QWidget):
             logger.debug(f'{mode} cache full, popping last widget: '
                          f'({oldest_widget})')
 
-            # typically this is fast enough, but if garbage collection is quick
-            # the widget may be deleted before we can delete it ourselves
-            try:
-                logger.debug(f'deleting widget from cache: {type(oldest_widget)}')
-                oldest_widget.setParent(None)
-                oldest_widget.deleteLater()
-            except RuntimeError:
-                pass
-
     def create_widget(self, item: TreeItem, mode: str) -> PageWidget:
         """Create the widget for ``item`` in ``mode``."""
         data = item.orig_data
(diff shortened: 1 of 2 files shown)


ZLLentz:
Good ol promotion of RuntimeError to segfault, a classic

tangkong:
I'm going to ask for a review as is.  In trying to make a test I founds some other oddities that prevent me from making a test that doesn't grab the guts of qt.  I'll try to address those in a separate PR.

PR #233, opened 2024-02-07
MNT: remove selectionModel calls, use `QTreeView.setCurrentIndex()`

### LINKED ISSUE #232: BUG: index in DualTree is not properly selected at init
## Expected Behavior
The QTreeView should select the top level index on init

## Current Behavior
We attempt to select the top-level item, and that is indeed the page the shows.  However inspection of the QTreeView reveals our selection is not propagating.  

```
(start atef config lfe.json)
select root first child: <atef.widgets.config.utils.TreeItem object at 0x7f6b60d2cac0>
-- show selected display: None
---> index below: None
sel_by_item: None
index selected at init: None

<Click on new item in tree view>

-- show selected display: at2l0
---> index below: blade states
-- show selected display: lfe demo
---> index below: at2l0
```

The window operates properly once we start interacting with it, but the initial state is a bit undefined.

## Possible Solution
???

 Make lots of print statements and inspect the treeview with something like: `print(DualTree.model.data(DualTree.tree_view.SelectionModel().currentIndex(), 0))`

## Context
This stymied my attempts to make a nice test

## Your Environment
pcds-5.8.1

tangkong:
## Description
Replaces selectionModel calls, use `QTreeView.setCurrentIndex()`.  

## Motivation and Context
Closes #232 , but also allows me to write (slightly) better tests for #231 

## How Has This Been Tested?
Interactively, but also I added a test.

This PR

## Pre-merge checklist

### DIFF
--- a/atef/widgets/config/window.py
+++ b/atef/widgets/config/window.py
@@ -519,9 +519,6 @@ class DualTree(DesignerDisplay, QWidget):
 
         self.toggle = Toggle()
 
-        # select top level root
-        self.tree_view.setCurrentIndex(self.model.index(0, 0, QtCore.QModelIndex()))
-
     def assemble_tree(self) -> None:
         """init-time tree setup.  Sets the tree into edit mode"""
         # self.tree_view = QtWidgets.QTreeView()
@@ -574,6 +571,9 @@ class DualTree(DesignerDisplay, QWidget):
         self.tree_view.selectionModel().selectionChanged.connect(
             self.show_selected_display
         )
+
+        # select top level root
+        self.tree_view.setCurrentIndex(self.model.index(0, 0, QtCore.QModelIndex()))
 
     def select_by_item(self, item: TreeItem) -> None:
         """Select desired TreeItem(and show corresponding page) in TreeView"""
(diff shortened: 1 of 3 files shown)


tangkong:
Before I go about fixing more bugs that didn't exist last year, let's get this through the pipes

atef/widgets/config/window.py
```diff
@@ -519,9 +519,6 @@ def __init__(
 
         self.toggle = Toggle()
 
-        # select top level root
-        self.select_by_item(self.root_item.child(0))
```
ZLLentz, line 523: Did this init step never actually need to be done then?

atef/widgets/config/window.py
```diff
@@ -556,9 +553,8 @@ def refresh_model(self) -> None:
 
         # Clear widget caches
         for cache in (self.edit_widget_cache, self.run_widget_cache):
-            for widget in cache.values():
-                widget.setParent(None)
-                widget.deleteLater()
+            for _ in range(len(cache)):
+                cache.popitem()
```
ZLLentz, line 557: Is this a continuation of the "let python do the garbo collection" mantra?

ZLLentz (approved):
Two small questions but it seems correct

atef/widgets/config/window.py
tangkong: yes it is.  Ran into a similar problem and applied the same solution

atef/widgets/config/window.py
tangkong: It actually gets moved to `DualTree.refresh_model()`, just in a slightly different form.

PR #235, opened 2024-02-09
BUG: Fix widget-deleted errors in `MultiModeValueEdit`

### LINKED ISSUE #234: BUG: Attempting to access deleted QComboBox in MultiModeValueEdit
## Expected Behavior
No broken widgets

## Current Behavior
Calls to EPICS can take a while to timeout, delaying callbacks from firing.  If we very rapidly change pages, widgets can be deleted before the callback completes, resulting in the classic runtime error C/C++ object deleted

```
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.8.1/bin/atef", line 10, in <module>
    sys.exit(main())
  File "/cds/home/r/roberttk/devrepos/atef/atef/bin/main.py", line 100, in main
    func(**kwargs)
  File "/cds/home/r/roberttk/devrepos/atef/atef/bin/config.py", line 94, in main
    app.exec()
  File "/cds/home/r/roberttk/devrepos/atef/atef/widgets/config/utils.py", line 2113, in fill_enums
    self.enum_input.addItem(text)
RuntimeError: wrapped C/C++ object of type QComboBox has been deleted
```

## Possible Solution
This might simply require catching the exception in the callback?  Or setting up the BusyCursorThread to hook something different on RuntimeErrors?  

1.  Open a config file with an enum comparison
2.  up and down arrow through the tree rapidly passing over the enum-comparison

## Context
Messing around with the GUI

## Your Environment
pcds-5.8.1

tangkong:
## Description
Makes the `BusyCursorThread` called in `MultiModeValueEdit` more tolerant of widgets being garbage collected while EPICS is timing out

## Motivation and Context
closes #234 

In summary, aspects of `MultiModeValueEdit` (`MMVE`) call EPICS, and if we're thumbing quickly through the pages, we can pop the parent page from the cache and delete it before the connection times out.   The callback would then fire, and cause ATEF to throw an error. 

We just catch those and ignore it if this is the case.

## How Has This Been Tested?
interactively, and by removing the cache limitation in `test_open_all_pages`

This PR

## Pre-merge checklist

### DIFF
--- a/atef/widgets/config/utils.py
+++ b/atef/widgets/config/utils.py
@@ -2109,17 +2109,22 @@ class MultiModeValueEdit(DesignerDisplay, QWidget):
                             enums_in_order.append(enum_str)
 
         def fill_enums():
-            for text in enums_in_order:
-                self.enum_input.addItem(text)
-            value = str(self.value.get())
-            if value in enums_in_order:
-                self.enum_input.setCurrentText(value)
+            try:
+                for text in enums_in_order:
+                    self.enum_input.addItem(text)
+                value = str(self.value.get())
+                if value in enums_in_order:
+                    self.enum_input.setCurrentText(value)
 
-            if set_mode:
-                if enums_in_order:
-                    self.set_mode(EditMode.ENUM)
-                else:
-                    self.set_mode(EditMode.STR)
+                if set_mode:
+                    if enums_in_order:
+                        self.set_mode(EditMode.ENUM)
+                    else:
+                        self.set_mode(EditMode.STR)
+            except RuntimeError:
+                # Widget sometimes destroyed before this completes
+                # simply return if this happens
+                return
 
         self.thread_worker = BusyCursorThread(func=get_signal_enums)
         self.thread_worker.task_finished.connect(fill_enums)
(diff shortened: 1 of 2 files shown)


tangkong:
A couple tests failed on the test I edited.  These tests passed locally, though occasionally segfaulted at the end of the pytest session.  I suspect there's another similar issue with flipping through all the pages quickly

tangkong:
I give up on testing with a small widget cache for now (reverted the commit).  qt tests just seem cursed.  Local testing passes with a singular passthrough, but with repeats we get segfaults from other tests.  (Why does expanding the cache let us pass those tests?  beats me)

The one through-line is that the failures are all thread / ophyd related.  It's possible there are other threaded operations running into problems similar to those this PR is trying to fix.  Why these errors are sometimes RuntimeErrors and sometimes segfaults gives me pause, but for now I think this is a reasonable addition.

ZLLentz (approved):
I wish this wasn't a generic `RuntimeError`, it'd be really nice to able to more broadly ignore and drop specifically the "widget is gone" and "slot/signal is gone" sorts of issues.

tangkong:
I looked into that, but aside from inspecting the exception message itself, I didn't see a nice way of determining the sub-type of the error.  I suppose we could do that but that seems brittle.

ZLLentz:
Yeah, I'm lamenting the implementation of pyqt here- this would have been a great time for them to define a specific error time for downstream consumers. I'm also against parsing the error message, etc., it's too brittle as you say.

PR #238, opened 2024-05-25
ENH: Add Template Configuration and Step

tangkong:
## Description
Adds a configuration/step for finding/replacing objects in an existing checkout, and running it.  GUI elements will be added in a separate effort

## Motivation and Context
A frequently requested feature.  This is the first part of this

Much refactoring needed to be done here, for a couple of reasons:
- The existing `FindReplaceAction` was not serializable
  - The callables used to encapsulate replacement actions could not be stored in a json
  - I created a more specific, but less flexible version (`RegexFindReplace`) that gets converted to the existing `FindReplaceAction` before application.
- A bunch of refactoring was needed to still type hint while preventing circular imports.  

Fun note: using `asyncio.run(config.compare())` in the test suite seemed to sometimes "kill the main thread", and would cause later bluesky plan tests to fail (which also wanted to access the "main thread" for the RunEngine.  This was resolved with `@pytest.mark.asyncio`, but in returning to the issue I wasn't able to reproduce the error

## How Has This Been Tested?
Interactively, primarily.  A small test was added, but more should come.

This PR.

## Pre-merge checklist

### DIFF
--- a/atef/config.py
+++ b/atef/config.py
@@ -316,7 +316,7 @@ class ToolConfiguration(Configuration):
 class TemplateConfiguration(Configuration):
     """
     A Configuration for applying changes to a template file.
-    Somewhat breaks the convention for Configuration by not directly having any
+    Breaks the convention for Configuration by not directly having any
     comparisons, but the generated file will have comparisons
     """
     # Path to a valid ConfigurationFile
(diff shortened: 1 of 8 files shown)


ZLLentz:
I think your PR description is out of date (you've done your to-dos)

ZLLentz (approved):
I think this is a reasonable implementation and I didn't find anything to nitpick.
I suspect it will be easier to try out once there are some basic ui elements associated with these.
I imagine that a prepared template in the config widgets will resolve to a group config in the prepared widgets? Should be neat.

I think regex is a good call, I wonder if some users will prefer a simpler templating though- like the shell-style templates provided by string.Template

tangkong:
> I imagine that a prepared template in the config widgets will resolve to a group config in the prepared widgets? Should be neat.

That is the intention.  I'm hoping to show something similar to the passive-checkout-in-active-checkout view
image

> I think regex is a good call, I wonder if some users will prefer a simpler templating though- like the shell-style templates provided by string.Template

Past me was definitely over optimistic about people using the guts of atef to write their own checkouts and write clever search / replace functions.  `string.Template` is definitely an interesting thought, we can change the default down the line.  I figured that without changing any settings, the "regex" option here reduces to a simple find/replace on strings.

PR #239, opened 2024-05-30
ENH/REF: Modify FillTemplatePage to have staging area

tangkong:
## Description
- Adds a staging area to `FillTemplatePage`, to be more clear about which changes are being queued for application
- Refactors various helpers and widgets to support this
- Adds tree-view that highlights closest tree-item when a row detail is examined

## Motivation and Context
In looking into reusing `FillTemplatePage` for templated checkouts, it became clear that the page needed a refactor
- If we were to serialize actions, we would need some way to display those queued actions.
  - Previously, `FillTemplatePage` would maintain an internal "all actions" list and a user could apply them gradually.  This made the current state of the loaded checkout ambiguous (had we applied changes yet?)
- Edit actions (`FindReplaceAction`) saved in a templated checkout would need to be loaded without knowledge of the search query that produced it.
  - Prior to this `FillTemplatePage` would gather all actions in any populated edit when asked to apply all.  The user would either have to know that these were the right actions or click through them all to verify
  - Now the staging area makes clear what will be applied when we save the file, and edits are not applied until the new file is saved

## How Has This Been Tested?
Entirely interactively so far.  But the test suite passes

This PR.

image

image

Compare to before:
<details><summary>Details</summary>
<p>

image

</p>
</details> 

## Pre-merge checklist

### DIFF
--- a/atef/widgets/config/find_replace.py
+++ b/atef/widgets/config/find_replace.py
@@ -386,6 +386,7 @@ class FillTemplatePage(DesignerDisplay, QtWidgets.QWidget):
 
     stage_all_button: QtWidgets.QPushButton
     open_button: QtWidgets.QPushButton
+    top_open_button: QtWidgets.QPushButton
     save_button: QtWidgets.QPushButton
     verify_button: QtWidgets.QPushButton
 
@@ -408,12 +409,15 @@ class FillTemplatePage(DesignerDisplay, QtWidgets.QWidget):
         self._devices: List[str] = []
         self.busy_thread = None
         self._partial_slots: list[WeakPartialMethodSlot] = []
+        self.file_name_label.hide()
+        self.type_label.hide()
         if filepath:
             self.open_file(filename=filepath)
         self.setup_ui()
 
     def setup_ui(self) -> None:
         self.open_button.clicked.connect(self.open_file)
+        self.top_open_button.clicked.connect(self.open_file)
         self.save_button.clicked.connect(self.save_file)
         self.stage_all_button.clicked.connect(self.stage_all)
         self.verify_button.clicked.connect(self.verify_changes)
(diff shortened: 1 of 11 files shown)


atef/widgets/config/run_base.py
```diff
@@ -430,7 +430,7 @@ def __init__(self, *args, **kwargs):
 
 def create_tree_from_file(
     data: Union[ConfigurationFile, ProcedureFile],
-    prepared_file: Union[PreparedFile, PreparedProcedureFile]
+    prepared_file: Optional[Union[PreparedFile, PreparedProcedureFile]] = None,
```
ZLLentz, line 433: The associated docstring for this is now out of date

atef/widgets/config/find_replace.py
```diff
@@ -56,7 +58,7 @@ def verify_file_and_notify(
     bool
         the verification success
     """
-    verified, msg = file.verify()
+    verified, msg = file.validate()
```
ZLLentz, line 61: Was this unknowingly broken for a few days?

tangkong: Yes 😢

tangkong:
> I always find it really hard to review atef

I'm not sure how to make this better.  atef is complicated but I think the web of signals and slots tends to make things opaque.

I've often wondered what it would take to make the atef development process more clear...

ZLLentz:
Design thoughts and nitpicks:
- The "select file" button is surprisingly hard to find from a blank template fill because it's at the bottom of a very long screen and selecting a file ends up populating the very top portion of the screen. I wonder if there should be another way to open a file, some clicky click magic in one of the top areas, for example.
- What is the intent of the "signals" column in the top-right view? It certainly isn't showing device signals, is it just for PV configs? If it's for PV configs I suggest marking the column as e.g. PVs instead of as Signals to evoke the same terminology as used in the rest of atef.
- I'm glad you have the movable dividers because qt default resizing is pain as always. One of these is slightly awkward though because it's so so close to the "Stage All" button, it's easy to click the button by mistake while aiming for the divider.
- I didn't see any bugs
- The "edits" table header is the only text element on the display that isn't capitalized (should probably be "Edits")

tangkong:
"no obvious bugs" has to be an all-time high for me 😆

ZLLentz (approved):
I'm :+1: on this design update

tangkong:
Great, thanks!  I'm going to merge this and forge ahead with the GUI elements that fold this into the main checkouts.

PR #248, opened 2024-08-01
ENH: add report output option to atef check cli

tangkong:
## Description
Adds the option to save a report after running a passive checkout from the CLI

## Motivation and Context
I promised this in a meeting earlier today and it didn't exist

## How Has This Been Tested?
Interactively: `atef check checkout.json -r report.pdf`

This PR, and in some docstrings

## Pre-merge checklist

### DIFF
--- a/atef/bin/check.py
+++ b/atef/bin/check.py
@@ -20,6 +20,7 @@ from ..check import Comparison, Severity
 from ..config import (AnyConfiguration, AnyPreparedConfiguration,
                       ConfigurationFile, FailedConfiguration,
                       PreparedComparison, PreparedFile, PreparedGroup)
+from ..report import PassiveAtefReport
 from ..result import Result
 from ..util import ophyd_cleanup
 
@@ -128,6 +129,11 @@ def build_arg_parser(argparser=None):
         "-p", "--parallel",
         action="store_true",
         help="Acquire data for comparisons in parallel",
+    )
+
+    argparser.add_argument(
+        "-r", "--report-path",
+        help="Path to the report save path, if provided"
     )
 
     return argparser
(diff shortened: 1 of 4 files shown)


atef/bin/check.py
```diff
@@ -130,6 +131,11 @@ def build_arg_parser(argparser=None):
         help="Acquire data for comparisons in parallel",
     )
 
+    argparser.add_argument(
+        "-r", "--report-path",
+        help="Path to the report save path, if provided"
+    )
```
ZLLentz, line 137: Have we ever talked about having a config file for atef? If we ever get around to something like that, a default reports directory would be a useful configurable (rather than remembering to -r each time)

ZLLentz (approved):
I'm :+1: on this with some thoughts and guesses about what the users will want next

atef/bin/check.py
```diff
@@ -452,6 +463,17 @@ async def check_and_log(
...
+    return prepared_file
+
+
+def save_report(prep_file: PreparedFile, report_path: str):
+    # Normalize report path
+    from pathlib import Path
+    save_path = Path(report_path).resolve()
+
+    doc = PassiveAtefReport(str(save_path), config=prep_file)
+    doc.create_report()
```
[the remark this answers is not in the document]
tangkong: Same, and I'm disappointed in myself.  In the past we've had to make a choice between active and passive, but here we don't.  Not that a helper couldn't do this.  It's probably worth refactoring this

atef/bin/check.py
tangkong: We haven't, all relevant information has been contained within the config file up until now.

A global config file could be helpful, though I'm not sure what other information we'd want to hold.

ZLLentz:
This is good to go if you didn't have anything else to add imo

tangkong:
I'll add some issues to track the thoughts you had here and merge this then 👍

PR #254, opened 2024-09-13
BUG: store _setpoint_pv in Target where possible

tangkong:
## Description
when selecting a ophyd device/signal, the original behavior would store both the PV name and device-attribute pair.  If the device signal has a separate write pv, we weren't saving that properly

Adjust this to save the `_setpoint_pv` if available, then the `pvname` after.  (ophyd signals have no write_pv attribute explicitly 🤷 )

## Motivation and Context
A bug Christian found while using atef and templates

## How Has This Been Tested?
Interactively

This PR 

## Pre-merge checklist

### DIFF
--- a/atef/widgets/config/data_active.py
+++ b/atef/widgets/config/data_active.py
@@ -874,7 +874,7 @@ class TargetEntryWidget(DesignerDisplay, QtWidgets.QWidget):
         dot_attr = attr.attr
         _attr = '_' + dot_attr.replace('.', '_')
         dev_name = full_name[:-len(_attr)]
-        write_pv = getattr(attr.signal, 'setpoint_pvname', attr.pvname)
+        write_pv = attr.setpoint_pvname or attr.pvname
 
         return Target(device=dev_name, attr=dot_attr, pv=write_pv)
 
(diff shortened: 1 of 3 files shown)


atef/widgets/config/data_active.py
```diff
@@ -874,8 +874,9 @@ def attr_to_target(self, attr: OphydAttributeData) -> Target:
         dot_attr = attr.attr
         _attr = '_' + dot_attr.replace('.', '_')
         dev_name = full_name[:-len(_attr)]
+        write_pv = getattr(attr.signal, 'setpoint_pvname', attr.pvname)
```
ZLLentz, line 877: Is it prohibitive to include the `setpoint_pvname` as part of `OphydAttributeData`?
https://github.com/pcdshub/atef/blob/ed137620cfd8f411e16fd27a68d806708b61512e/atef/widgets/ophyd.py#L67-L76

ZLLentz: This isn't necessary to merge the PR but it's something I noticed because it seems weird to need to dig into the `attr.signal` at this stage.

atef/widgets/config/data_active.py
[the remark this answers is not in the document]
tangkong: whoops yes

atef/widgets/config/data_active.py
tangkong: It might take a bit of refactoring but it's probably clearer moving forward.  The existing ophyd device model does a similar access and we'd just need to clean that up

ZLLentz (approved):
lgtm

PR #256, opened 2024-09-17
PERF: refactor atef.bin cli entrypoint to defer imports

tangkong:
## Description
- takes a knife to `atef.bin` to defer imports until absolutely necessary

In more detail:
- moves all `build_arg_parser` type functions to a module separate from the subcommand logic
- `atef.bin.main` now grabs those parsers to create the top-level parser, avoiding the meaty functional imports
- the `main` sub-commands are given an additional layer of wrapping, and `atef.bin.main:main` now calls a partial function to get the `atef.bin.<subcommand>:main` function, which then gets fed the parsed kwargs.

## Motivation and Context
I keep vendoring this cli hook and never improving it.  This was more of a proof to myself that it could work.  If we like it I can start using it elsewhere or even maybe cookiecutter it.

I should really work on more important tasks but I refactored it in my brain during a post-dinner walk and wanted to put it to paper

## How Has This Been Tested?
Before
```bash
$ time atef config -h
...
real    0m8.487s
user    0m6.721s
sys     0m1.158s
```

After
```bash
$ time atef config -h
...
real    0m0.518s
user    0m0.184s
sys     0m0.075s
```
Aside from this, lots of interactive testing to make sure all the entrypoints are working.

This PR

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/256-perf_cli.rst
+++ b/docs/source/upcoming_release_notes/256-perf_cli.rst
@@ -1,0 +1,22 @@
+256 perf_cli
+############
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- improves the performance of the CLI entrypoint, defering functional imports as longas possible
+
+Contributors
+------------
+- tangkong
--- a/atef/bin/check.py
+++ b/atef/bin/check.py
@@ -3,7 +3,6 @@
 """
 from __future__ import annotations
 
-import argparse
 import asyncio
 import enum
 import itertools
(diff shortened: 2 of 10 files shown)


docs/source/upcoming_release_notes/256-perf_cli.rst
```diff
@@ -0,0 +1,22 @@
...
+--------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- improves the performance of the CLI entrypoint, defering functional imports as longas possible
```
ZLLentz, line 18:
```suggestion
- improves the performance of the CLI entrypoint, deferring functional imports as long as possible
```

ZLLentz (commented):
This seems reasonable to me and is a big speedup due to some of the slower imports in some of the code paths. My only in-line review comment is a typo check.

I have mixed feelings about the various variants/options for a subparser scheme like this one. The previous arrangement had some performance problems but functionally made clear separations between the implementations of the different subcommands. Part of me wonders if there was a way to keep the subparsers a little more separate without needing to change so much, something like:

`atef.bin.subcmd`
```
import argparse

def build_parser():
    ...

def main(**kwargs):
    from .subcmd_main import subcmd_main
    subcmd_main(**kwargs)
```

And then the overall structure doesn't change too much and we still import a bunch of subparsers but the heavy imports are deferred and the subparsers aren't collocated.

But maybe there are even cleaner ways as well.

This is all just musing. I'm happy to merge with the typo fixed.

tangkong:
Hmm I might like that formulation more.  I'll give that a whirl

tangkong:
superceded by #257

PR #259, opened 2024-10-30
MNT: increase timeout limits (limits on all form spinboxes), prevent aggressive navigation

tangkong:
## Description
- increase the limit on spinboxes generated from MultiInputFormDialog, in response to a request that set-value-step timeouts not be limited to 99s (we have very slow motors)
- navigate BACK to the parent page when creating a new check in SetValueStep
  - This negates the default behavior of StepPage, which selects new items when they are created. 
  - Because SetValueStep initialized a row/new item without setting a PV, this resulted in extraneous navigation back to the parent page to set the PV, then back to the details to set the comparison information
  - Hacky but I'm too lazy to figure out a better way

## Motivation and Context
Bug reports from @c-tsoi 

## How Has This Been Tested?
Interactively.  Test suite still passes

This PR

## Pre-merge checklist

### DIFF
--- a/atef/widgets/config/data_active.py
+++ b/atef/widgets/config/data_active.py
@@ -597,6 +597,7 @@ class SetValueEditWidget(DesignerDisplay, DataWidget):
             self.checks_table.add_row(data=check)
 
         self.checks_table.cellClicked.connect(self.update_all_desc)
+        self.checks_table.table_updated.connect(self.nav_to_self)
 
         # checkboxes
         self.bridge.halt_on_fail.changed_value.connect(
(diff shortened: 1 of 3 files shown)


atef/widgets/config/data_active.py
```diff
@@ -652,6 +653,14 @@ def update_all_desc(self, *args, **kwargs):
             row_widget: CheckRowWidget = self.checks_table.cellWidget(ind, 0)
             row_widget.update_summary()
 
+    def nav_to_self(self):
+        # A bit of a hack to prevent navigating to details prematurely
+        # normally an isinstance check would be good, but circular imports
+        try:
+            self.parent().parent().full_tree.select_by_data(self.data)
```
ZLLentz, line 660: Briefly scrutinizing this: why is the parent's parent always the instance that owns the full tree?

ZLLentz (approved):

atef/widgets/config/data_active.py
tangkong: The `SetValueStep` widget is inside a QWidget container, which is itself inside a `StepPage(PageWidget)` widget.  only those `PageWidget` classes can see the full tree

tangkong:
I forgot about this.  Merging now

Issue #44, opened 2022-04-19
Feature: atef config GUI Tree item context menus for adding/deleting/renaming/copying


klauer:
* Right click a device configuration and add a checklist underneath it
* Right click a checklist and add a comparison underneath it
* Right click a checklist/comparison and copy/paste it elsewhere

_Originally posted by @klauer in https://github.com/pcdshub/atef/pull/28#discussion_r853302927_

PR #266, opened 2025-07-07
Adding the ability to copy/paste comparisons and groups

### LINKED ISSUE #218: ENH: Copy / Paste from tree?
## Expected Behavior
It would be nice to be able to copy portions of a checkout and paste it into another.  Say if one active checkout had a large ProcedureGroup we wanted but none of the other bits, we could copy it specifically into another checkout.

This could be only valid for groups and comparisons, maybe?

This would be good for scripts that generate checkouts, so we could combine those into other bigger better checkouts.

## Current Behavior
There is no support for this currently.

## Possible Solution
I don't actually know how copy/pasting complex data structures works in other apps.  We could save the relevant dataclass to an internal "clipboard".  We'd then have to add some sort of paste method, which knows where to drop this dataclass inside a given group.  Maybe this requires a separate dialog?

Are there places where the paste location is ambiguous?  Are there configurations that have comparisons in two different places/collections?

Maybe this could also copy the serialized version for pasting into saved checkouts?  This seems easy but might be too confusing.

## Context
Thought about this while writing a script that might take a pmgr configuration and make a checkout.

## Your Environment
pcds-5.7.3, atef-1.2.0

KaushikMalapati:
## Description

Added a context window with the options to copy/paste at the selected item in the tree. 
Copying will make a deepcopy of the item
Pasting will throw an exception if nothing is pasted. If attempting to paste a Comparison when another Comparison is selected, it will attempt to add the comparison to the parent of the selected comparison.
Pasting non-Comparisons only work if a ConfigurationGroup or ProcedureGroup is added, and will add it to said group.
When pasting an item, if the page is already in the cache the add_[comparison/config]_row method is used to update the page object 

## Motivation and Context

Closes #218 and #44

## How Has This Been Tested?

Interactively by copy and pasting comparisons/Configurations/Procedures in the sample checkouts
I was also able to save and reopen after pasting a Comparison

image

## Pre-merge checklist

### DIFF
--- a/atef/widgets/config/window.py
+++ b/atef/widgets/config/window.py
@@ -325,24 +325,24 @@ class Window(DesignerDisplay, QMainWindow):
         # Make another deepcopy in case the data is pasted multiple times
         old_data = self.clipboard[1]
         self.clipboard = deepcopy(self.clipboard)
-        if isinstance(paste_data, ConfigurationGroup) and isinstance(self.clipboard[1], Configuration):
+        if isinstance(paste_data, ConfigurationGroup) and isinstance(old_data, Configuration):
             paste_data.configs.append(old_data)
             page.add_config_row(config=old_data)
-        elif isinstance(paste_data, ProcedureGroup) and isinstance(self.clipboard[1], ProcedureStep):
+        elif isinstance(paste_data, ProcedureGroup) and isinstance(old_data, ProcedureStep):
             paste_data.steps.append(old_data)
             page.add_config_row(config=old_data)
-        elif isinstance(self.clipboard[1], ComparisonToTarget) and isinstance(paste_data, SetValueStep):
+        elif isinstance(old_data, ComparisonToTarget) and isinstance(paste_data, SetValueStep):
             page.specific_procedure_widget.checks_table.add_row(data=old_data)
-        elif isinstance(self.clipboard[1], Comparison) and isinstance(paste_data, Configuration):
+        elif isinstance(old_data, Comparison) and isinstance(paste_data, Configuration):
             paste_data.shared.append(old_data)
             page.add_comparison_row(comparison=old_data)
-        elif isinstance(self.clipboard[1], Comparison) and isinstance(paste_data, AnyComparison):
+        elif isinstance(old_data, Comparison) and isinstance(paste_data, AnyComparison):
             page.specific_comparison_widget.add_comparison(comparison=old_data)
             page.specific_comparison_widget.update_comparison_list()
         else:
             # Either trying an invalid paste (e.g., ConfigurationGroup into Comparison)
             # or pasting support was nto added for this type yet
-            raise RuntimeError(f'Pasting {self.clipboard[1].__class__.__name__} into {paste_data.__class__.__name__} is not supported')
+            raise RuntimeError(f'Pasting {old_data.__class__.__name__} into {paste_data.__class__.__name__} is not supported')
 
     def context_menu(self, position):
         """Open context window"""
--- a/atef/tests/test_widgets.py
+++ b/atef/tests/test_widgets.py
@@ -305,7 +305,6 @@ def test_copypaste(qtbot: QtBot, config: os.PathLike):
             print(e)
         curr_index = new_index
 
-    # if there are epics calls they may still be going on at test teardown?...
     qtbot.addWidget(window)
 
 


tangkong:
> how would I check for threading issues
This checklist item is more a reminder for the author, there's no way to guarantee we're free from threading issues but you should do your best to make sure if you run blocking tasks they shouldn't be in the main execution thread (gui thread)

> and do you know how to get the context window widget to pass to qtbot for a test case?
Passing a widget to qtbot also handles all its children and often this is enough.  How are you simulating / inspecting the context menu?

tangkong:
I might avoid testing the context menu creation, and instead focus on testing the methods you've written that actually perform the actions.  Integration testing context menus might be cumbersome/brittle (you'll have to position the mouse, invoke qtbot mouse clicks, etc)

If there are signals that trigger on click, perhaps you just emit those manually?

atef/procedure.py
```diff
@@ -139,7 +139,7 @@ class SetValueStep(ProcedureStep):
     #: Only mark the step_result as successful if all actions have succeeded
     require_action_success: bool = True
 
-    def children(self) -> List[ComparisonToTarget]:
+    def children(self) -> List[Comparison]:
```
tangkong, line 142: Whoops, good catch

tangkong (commented):
I think this is really nifty.  I'm going to take it for a bit more of a in depth test run, but while reading over the code I was struck by all the special handling you had to set up for each step type. 

Might it make more sense to place that logic into the items themselves?  Perhaps we let each `Configuration`/`ProcedureStep`/`Comparison` know what kind of data makes sense to receive as a copy/paste, then just call something like `PVConfiguration.paste()` from the window method?

That might make this all a bit more maintainable, and also makes this logic usable if people ever mess around with the python objects.

tangkong:
Just one note from interactive testing.

* There are some cases where we we don't raise the useful `RuntimeError`
  * e.g.: Pasting a comparison into a ConfigurationGroup -> AttributeError: 'ConfigurationGroup' object has no attribute 'shared'

tangkong:
I just did a:
* right-click -> copy a comparison
* right-click-paste on ConfigurationGroup in the tree
image

tangkong:
That might be a nice simplification, but I fear we'll still end up with this extensive if-else tree that is tough to reason about.  

The less we have to rely on grabbing the internals of each widget (touching table views directly, etc) the better imo.  Whether or not that's feasible is to be determined, I need to refresh my atef memory.

atef/widgets/config/window.py
```diff
@@ -294,8 +300,102 @@ def _new_tab(
...
+        """Copy the currently selected item and its parent to the clipboard"""
+        current_item = self.get_current_tree().current_item
+        parent_data = current_item.parent().orig_data
+        item_data = current_item.orig_data
+        # make deep copies now in case items are modified after copying but before pasting
+        if isinstance(parent_data, SetValueStep):
+            item_data = parent_data.success_criteria[parent_data.children().index(item_data)]
+        self.clipboard = (deepcopy(parent_data), deepcopy(item_data))
+
+    def start_pasting(self, list_attr=None, location=None):
```
tangkong, line 317: On second read, this method is a bit confusing.  The return type is `Union[Tuple[PageWidget, AnyDataclass], AnyDataclass]`?  After which you have to inspect the return from this method to decide what to do?

The purpose of this method is to gather the data and paste location?  A good docstring here might be helpful

tangkong:
@KaushikMalapati Any chance you've got time to round this out this week?  We're looking to tag and it'd be cool to get this in.

Thinking about it a bit more, I'm onboard with your proposal above (limit pasting to selecting groups, use the page methods to handle gui and dataclass insertion)

tangkong:
Ok, that sounds reasonable.  

1: This could be useful, if we have the time to implement it.  If I recall, my goal was to minimize the amount of new work / widgets we had to build, so I tried to reuse the familiar (ideally) Comparison page.  I didn't build `ComparisonToTarget` to really be reused elsewhere, but it is worth considering consolidating these comparisons. In general we specify the "target" of a comparison in different ways across the different groups, which has been confusing to me in the past (though necessary to prevent repetition).

2: I like to use mappings as a sort of multi-use case statement.  If you're only using it here to replace one if-else block, maybe it's not so beneficial.  I imagine you'd be mapping widgets to the appropriate method, which you'd have to `getattr`?  

I'm not exactly sure how an abstract class would help here.  If you make all the other page widgets inherit from a class that requires a consistent method for adding a comparison, that could help but also add a bunch of unnecessary boilerplate.  I think the if-else block is fine for now, but if we end up wanting to re-use the logic we can consider refactoring.

3: The `by_attr` / `by_pv` / `shared` maps **are** required, to my understanding.  Those maps are used in serialization, and are read when we transition to run-mode.  It does seem like you're adding comparisons to `shared` though, which should work as long as we can change the "target" of the comparison after.

PR #274, opened 2025-11-14
ENH: Adding parent info to status output

ctslac:
## Description

I added parent step information to the status output so that users can tell which common component procedure group they are running and the corresponding step information.

## Motivation and Context

This is to help users with tracking which step they are on in the live gui.

## How Has This Been Tested?

Visually, not sure if I need to do aditional tests.

## Pre-merge checklist

### DIFF
--- a/atef/config_model/active.py
+++ b/atef/config_model/active.py
@@ -571,6 +571,9 @@ class PreparedProcedureStep:
 
     async def run(self) -> Result:
         """Run the step and return the result"""
+
+        self.parent.name = getattr(self.parent, "name", "(no parent)")
+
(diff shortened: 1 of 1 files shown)


tangkong:
Tests are failing because we started relying on the parent being set where sometimes it isn't.  I think it should be set in those cases, and we can/should fix that here.   It looks like the PreparedSteps are all missing this for some reason

ctslac:
Is the flow chart you showed me earlier reflective of where it should have been set?

tangkong:
I'm not sure which flowchart you're referring to here, but it's a good exercise to read through the code to try and understand where the parent might get set

PR #278, opened 2026-06-25
TST: tweak for updated pytest syntax parsing

ZLLentz:
## Description

Try to resolve the CI errors in this repo by updating the parameterize syntax

## Motivation and Context

A recent pytest update changed the meaning of the parameterize syntax to more closely match normal python syntax

## How Has This Been Tested?

It will be tested in CI here

Here only

## Pre-merge checklist

### DIFF
--- a/atef/tests/test_abort.py
+++ b/atef/tests/test_abort.py
@@ -18,7 +18,7 @@ def assert_task_running(tree: DualTree, is_running: bool = True):
             and tree.running_task.done() != is_running)
 
 
-@pytest.mark.parametrize("click_abort,", [
+@pytest.mark.parametrize("click_abort", [
     (True,),
     (False,),
 ])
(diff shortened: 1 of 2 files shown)


ZLLentz:
This is now failing on a test failure instead of a test suite failure- that's progress I guess. I'll try to understand what is happening here.

ZLLentz:
I'm intentionally omitting the pre-release docs because I plan to remove them from every repo over time- I guess I'll remove them from atef today

ZLLentz:
I think this is now the full fix needed here

KaushikMalapati (approved):
I kind of feel like pinning a pytest versions across the repos that use it to avoid this kind of inconvenience

ZLLentz:
It's certainly something we can do- for example, we can specify it in pcds-ci-helpers

The reason I'm opting to fix these is because the syntax was always _technically_ wrong- I perceive this as pytest having let us "get away" with these trailing commas that have a specific meaning in other python contexts.

